PR #147, opened 2023-05-09
Enhancement: Consistently use maps for timeline data

pleek91:
# Description
Replaces the existing array type with a map type by default. Previously an array was being converted into a map (in multiple places). Requiring a map be passed in prevents having to convert. Also fixes how the data is passed to the worker by using vue's `toRaw` utility so we can preserve the map as is and pass it to the worker.

### DIFF
--- a/src/FlowRunTimeline.vue
+++ b/src/FlowRunTimeline.vue
@@ -18,7 +18,6 @@
   } from 'vue'
   import { Guides } from '@/containers/guides'
   import {
-    GraphTimelineNode,
     nodeThemeFnDefault,
     TimelineThemeOptions,
     FormatDateFns,
@@ -42,6 +41,7 @@
     createTimeScale,
     nodeClickEvents
   } from '@/pixiFunctions'
+  import { TimelineData } from '@/types/timeline'
   import {
     getDateBounds,
     parseThemeOptions
@@ -51,7 +51,7 @@
   const animationThreshold = 500
 
   const props = defineProps<{
-    graphData: GraphTimelineNode[],
+    data: TimelineData,
     isRunning?: boolean,
     theme?: TimelineThemeOptions,
     formatDateFns?: Partial<FormatDateFns>,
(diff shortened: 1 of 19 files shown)


brandonreid (approved):
Super welcomed and well executed refinement!

src/utilities/time.ts
```diff
@@ -91,28 +92,31 @@ export function formatDate(date: Date): string {
...
+      maxEndTime = Math.max(maxEndTime, endTime)
+    } else {
+      maxEndTime = endTime
     }
+  })
 
-    return [min ?? startTime, max ?? endTime]
-  }, [null, defaultMaxEndTime])
-
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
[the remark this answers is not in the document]
pleek91: Sorry should have added an explanation to the comment. Typescript doesn't follow `let` very well sometimes. Here I couldn't get it to understand that it could in fact be null

src/pixiFunctions/timelineNodes.ts
```diff
@@ -265,22 +257,21 @@ export class TimelineNodes extends Container {
   /**
    * Update Functions
    */
-  public update(newData: GraphTimelineNode[]): void {
-    this.graphData = newData
```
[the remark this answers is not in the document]
pleek91: I'm not sure, but I tried to not change any functionality with how things worked. We can probably clean this up at some point.

src/FlowRunTimeline.vue
```diff
@@ -390,11 +390,11 @@
       }
     })
 
-    watch(() => props.graphData, () => {
```
brandonreid, line 393: This may not be getting called, when viewing the UI, the "slow feed data" option is no longer slow feeding nodes. I'm thinking updated data is not registering somewhere.

pleek91: looks like the watcher is being called and the layout is trying to update but the worker loses the layout setting 🤔

src/workers/nodeLayout.worker.ts
```diff
@@ -53,41 +53,16 @@ onmessage = async ({
...
-
-    if (timeScale && (layoutSetting || graphDataStore !== newData)) {
-      graphDataStore = prepareGraphData(newData)
-      await calculateNodeLayout()
-      const response: NodeLayoutWorkerResponseData = {
-        layout,
-        centerViewportAfter,
-      }
-      postMessage(response)
+  if (data && timeScale && currentLayoutSetting) {
```
brandonreid, line 56: @pleek91 I updated this to check for `currentLayoutSetting` rather than a `layoutSetting` being passed in, since when data updates we don't pass the layout in again. Fixed the problem we had that was preventing us from merging this PR. I'ma merge it!!

src/FlowRunTimeline.vue
brandonreid: Added a comment about how this was fixed. We good!

PR #179, opened 2023-08-15
update to v2 styles

brandonreid:


src/pixiFunctions/nodeSprites.ts
```diff
@@ -15,6 +15,7 @@ let nodeBoxTextureCache: Map<number, BoxTextures> | undefined
 let arrowTextureCache: Map<number, Texture> | undefined
 let roundedBorderRectCache: Map<RoundedBorderRectCacheKey, BorderRectTextures> | undefined
 
+const transparent = 0x00000000
```
znicholasbrown, line 18: What's "0x00000000"? I'm not sure how to think about this variable

brandonreid: Ah true, huh.. It just needs to be a unique hex number. Technically that's a hexadecimal color with zero transparency (first 2 numbers after the x), although that hex value results in black as the transparency channel isn't supported.

brandonreid: I'll leave a comment in there

src/utilities/style.ts
```diff
@@ -34,8 +34,8 @@ export function parseThemeOptions(overrides?: ThemeStyleOverrides): ParsedThemeS
     spacingGuideLabelPadding: spacingToNumber(overrides?.spacingGuideLabelPadding ?? '4px'),
     spacingPlayheadGlowPadding: spacingToNumber(overrides?.spacingPlayheadGlowPadding ?? '8px'),
     spacingPlayheadWidth: spacingToNumber(overrides?.spacingPlayheadWidth ?? '2px'),
-    borderRadiusNode: spacingToNumber(overrides?.borderRadiusNode ?? '8px'),
-    borderRadiusButton: spacingToNumber(overrides?.borderRadiusButton ?? '8px'),
+    borderRadiusNode: spacingToNumber(overrides?.borderRadiusNode ?? '6px'),
+    borderRadiusButton: spacingToNumber(overrides?.borderRadiusButton ?? '6px'),
```
znicholasbrown, line 38: Given that we've got tokens for spacing, should we still be hardcoding these sorts of defaults? i thought that was what the tokens were good for

znicholasbrown (approved):
a few questions, none blocking

src/pixiFunctions/nodeSprites.ts
znicholasbrown: Got it - thanks for the explanation!

src/utilities/style.ts
brandonreid: We do have `--p-spacing-radius-default`, that will be passed in from the consumer, as the setup is designed currently, so this is just a local default.

On a side note, since we'd be using the 8px grid (pretty ubiquitous out there these days) and tailwind uses that out of the box, we've opted to keep using tailwind styles for spacing. So only a few specific spacing tokens will exist, like the default radius, focus ring offset, etc.

znicholasbrown: Gotcha, makes sense. Why the change to 6px here then?

brandonreid: Just so it looks on brand in the demo I guess.. I suppose falling back to a value we're likely to actually use makes sense too?

I do think, since we have semantic tokens like this now, it may be worth revisiting all this theming stuff and just letting the component itself utilize the tokens, rather than the consumer passing them in (other than the node styles). Idk.

PR #209, opened 2023-09-27
Add new models

pleek91:
# Description
Adds models to match the new endpoint being built.

### DIFF
--- a/src/components/FlowRunGraph.vue
+++ b/src/components/FlowRunGraph.vue
@@ -5,7 +5,9 @@
 </template>
 
 <script lang="ts" setup>
+  import { FlowRunGraphConfig } from '@/models/FlowRunGraph'
+
   defineProps<{
-    flowRunId: string,
+    config: FlowRunGraphConfig,
   }>()
 </script>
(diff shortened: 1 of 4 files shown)


src/components/FlowRunGraph.vue
```diff
@@ -5,7 +5,9 @@
 </template>
 
 <script lang="ts" setup>
+  import { FlowRunGraphConfig } from '@/models/FlowRunGraph'
+
   defineProps<{
-    flowRunId: string,
+    config: FlowRunGraphConfig,
```
pleek91, line 11: This is noteworthy to call out. By accepting the `flowRunId` and a `fetch` method to get the graph data. This component can do its own fetching and refreshing of data. Which means it can fetch its own sub flow data rather than depending on the implementation to listen for events. 

I'm not sure how this will effect the implementation forcing an update. Might expose a `refresh` method on the component instance itself.

brandonreid: Ah I dig this! The model binding hurdles were def a bit much there.

brandonreid (approved):

PR #211, opened 2023-09-27
Run pixi in a web worker

pleek91:
# Description
Moves pixi itself into a worker. 

Learning from
https://www.npmjs.com/package/@pixi/webworker
https://github.com/bigtimebuddy/pixi-webworker-example
https://pixijs.com/examples/offscreen-canvas/basic

### DIFF
--- a/src/components/RunGraph.vue
+++ b/src/components/RunGraph.vue
@@ -1,13 +1,35 @@
 <template>
-  <div class="run-graph">
-    hello world
-  </div>
+  <canvas ref="canvas" />
 </template>
 
 <script lang="ts" setup>
+  import { onMounted, ref } from 'vue'
   import { RunGraphConfig } from '@/models/RunGraph'
+  // eslint-disable-next-line import/default
+  import PixiWorker from '@/workers/worker.ts?worker'
 
   defineProps<{
     config: RunGraphConfig,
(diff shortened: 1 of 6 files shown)


src/components/RunGraph.vue
```diff
@@ -1,13 +1,35 @@
...
+  onMounted(() => {
+    if (!canvas.value) {
+      throw new Error('Canvas does not exist')
+    }
+
+    const view = canvas.value.transferControlToOffscreen()
+    const worker = new PixiWorker()
+
+    // this isn't type safe
+    worker.postMessage({
```
znicholasbrown, line 26: If you'll be communicating frequently with the worker, it might be helpful to create an intermediary function that _is_ typesafe like we do [here](https://github.com/PrefectHQ/nebula-ui/blob/main/src/workers/util/promise-channel.ts) - certainly some room for error but it could save you debug time

znicholasbrown (approved):

src/components/RunGraph.vue
pleek91: Yeah I'll implement something like that probably. Though I'm not sure yet if or how much communication will go back and forth. I'm sure there will be some pixi events that will need to be emitted through the component.

PR #213, opened 2023-09-28
Revert to using an in process pixi application

pleek91:
# Description
After discovering FontFaceObserver and pixi-worker won't run in a worker I'm reverting back to running pixi in process while preserving the offscreen canvas. 

Also keeping the worker which will be repurposed for layout calculations.


znicholasbrown (approved):
Sad 😔

PR #214, opened 2023-09-29
Add viewport and organize how pixi objects will get created and shared

pleek91:
# Description
Gets the viewport wired up

Also trying out a pattern for organizing the various pixi objects. Previously we had a shared state object and returning/passing a lot of stuff around. This takes advantage of exporting `let` variables which cannot be written to when they are imported but can be written to in the same file. 

Also created a `start` method (name will probably change) that will be responsible for running the app and a `stop` method for shutting it down.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
   },
   "dependencies": {
     "date-fns": "2.30.0",
+    "pixi-viewport": "^5.0.2",
     "pixi.js": "^7.3.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
   },
   "dependencies": {
     "date-fns": "2.30.0",
-    "pixi-viewport": "^5.0.2",
-    "pixi.js": "^7.3.1"
+    "pixi-viewport": "5.0.2",
+    "pixi.js": "7.3.1"
   }
 }
--- a/src/components/RunGraph.vue
+++ b/src/components/RunGraph.vue
@@ -3,9 +3,9 @@
 </template>
 
 <script lang="ts" setup>
-  import { onMounted, ref } from 'vue'
+  import { onMounted, onUnmounted, ref } from 'vue'
   import { RunGraphConfig } from '@/models/RunGraph'
-  import { start } from '@/objects'
+  import { start, stop } from '@/objects'
   import { WorkerMessage, worker } from '@/workers/runGraph'
 
   defineProps<{
@@ -34,4 +34,6 @@
 
     start(stage.value)
   })
+
+  onUnmounted(() => stop())
 </script>
--- a/src/components/RunGraph.vue
+++ b/src/components/RunGraph.vue
@@ -1,18 +1,18 @@
 <template>
-  <canvas ref="canvas" />
+  <div ref="stage" />
 </template>
 
 <script lang="ts" setup>
   import { onMounted, ref } from 'vue'
   import { RunGraphConfig } from '@/models/RunGraph'
-  import { createApplication } from '@/objects/application'
+  import { start } from '@/objects'
   import { WorkerMessage, worker } from '@/workers/runGraph'
 
   defineProps<{
     config: RunGraphConfig,
   }>()
 
-  const canvas = ref<HTMLCanvasElement>()
+  const stage = ref<HTMLDivElement>()
 
   worker.onmessage = onMessage
 
@@ -28,16 +28,10 @@
   }
 
   onMounted(() => {
-    if (!canvas.value) {
+    if (!stage.value) {
       throw new Error('Canvas does not exist')
     }
 
-    const view = canvas.value.transferControlToOffscreen()
-
-    createApplication(view)
-
-    if (process.env.NODE_ENV === 'development') {
-      (globalThis as any).__PIXI_STAGE__ = canvas.value
-    }
+    start(stage.value)
   })
 </script>
--- a/src/objects/application.ts
+++ b/src/objects/application.ts
@@ -1,10 +1,19 @@
 import { Application } from 'pixi.js'
+import { stage } from '@/objects/stage'
 
 export let application: Application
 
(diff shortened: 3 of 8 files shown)


src/objects/viewport.ts
```diff
@@ -0,0 +1,30 @@
...
+      friction: 0.9,
+    })
+
+  application.stage.addChild(viewport)
+
+  // add a red box
+  const sprite = viewport.addChild(new Sprite(Texture.WHITE))
+  sprite.tint = 0xff0000
+  sprite.width = sprite.height = 100
+  sprite.position.set(100, 100)
```
pleek91, line 29: This is just temporary to put something on the viewport so you can see it working

package.json
```diff
@@ -59,6 +59,7 @@
   },
   "dependencies": {
     "date-fns": "2.30.0",
+    "pixi-viewport": "^5.0.2",
     "pixi.js": "^7.3.1"
```
znicholasbrown, line 63: Both of these probably need to be pinned

src/components/RunGraph.vue
```diff
@@ -1,18 +1,18 @@
 <template>
-  <canvas ref="canvas" />
+  <div ref="stage" />
```
znicholasbrown, line 2: big 🕺🏻 energy

src/objects/application.ts
```diff
@@ -1,10 +1,19 @@
...
 
-export function createApplication(view: OffscreenCanvas): void {
+export function createApplication(): void {
   application = new Application({
-    view,
     background: '#1099bb',
+    resizeTo: stage,
   })
+
+  stage.appendChild(application.view as HTMLCanvasElement)
```
znicholasbrown, line 12: I'm guessing this casting is necessary cause pixi doesn't have types for this?

src/objects/stage.ts
```diff
@@ -0,0 +1,5 @@
+export let stage: HTMLDivElement
```
znicholasbrown, line 1: this is clever, I didn't realize this worked

src/objects/viewport.ts
```diff
@@ -0,0 +1,28 @@
...
+    events: application.renderer.events,
+    passiveWheel: false,
+  })
+
+  viewport
+    .drag()
+    .pinch()
+    .wheel()
+    .decelerate({
+      friction: 0.9,
```
znicholasbrown, line 18: For my own knowledge, is there some significance to the friction value?

src/objects/viewport.ts
znicholasbrown: i see it working

znicholasbrown (approved):

znicholasbrown:
It's perhaps intentional or a consequence of a pixi default but the demo is using some kind of acceleration on mouse movements that means movement of mouse:picture in the canvas isn't 1:1 - it's probably fine? I can't tell if it's a good experience or a missed expectation, just calling it out

package.json
znicholasbrown: though honestly don't need to be during development, now that i think about it

src/objects/viewport.ts
pleek91: This was a value carried over from the previous version. Without some friction if you release while moving everything just flies off screen.

znicholasbrown: it occurs to me now that my comment on the PR is probably related to this value, since this is passed to a function called "decelerate" 🤦🏻

src/objects/application.ts
pleek91: It has a type but it's not compatible with `Node`. They supposedly fixed this in 7.0.3 but didn't seem to work still. 

From my experience almost all of the methods that accept a node aren't happy with perfectly valid elements. So casting here seamed fine.

package.json
pleek91: Agreed, will pin in the morning

src/objects/stage.ts
pleek91: Seems to work nicely. We'll see if it causes me any headaches.

src/objects/application.ts
znicholasbrown: yeah fair enough, d3 often has this problem as well

pleek91:
> It's perhaps intentional or a consequence of a pixi default but the demo is using some kind of acceleration on mouse movements that means movement of mouse:picture in the canvas isn't 1:1 - it's probably fine? I can't tell if it's a good experience or a missed expectation, just calling it out

Does this match what you're seeing on the current graph in staging or prod? My sim is to not change how that feels. There were more physics settings before but everything but the friction seamed to be the default values.

znicholasbrown:
> > It's perhaps intentional or a consequence of a pixi default but the demo is using some kind of acceleration on mouse movements that means movement of mouse:picture in the canvas isn't 1:1 - it's probably fine? I can't tell if it's a good experience or a missed expectation, just calling it out
> 
> Does this match what you're seeing on the current graph in staging or prod? My sim is to not change how that feels. There were more physics settings before but everything but the friction seamed to be the default values.

It's a little tough to tell tbh - no? but maybe because the demo is such a simpler representation, it's possible the scale is what makes the feeling different.

Discovered a fun bug while checking though, not sure when this got introduced but maybe not worth addressing right now? (second pic is to confirm it's present without a CVD setting present)

Screenshot 2023-09-29 at 12 53 00 AM
Screenshot 2023-09-29 at 12 51 40 AM

package.json
pleek91: Updated

pleek91:
@brandonreid do you know whats up with the screenshots above?

brandonreid:
> @brandonreid do you know whats up with the screenshots above?

That's a known issue in Safari. I'm hoping the update to v7 resolves this. I have a ticket logged for it in the graphs repo. Essentially the sprites are rendering pink where there should be transparent values, again only in Safari.

PR #221, opened 2023-10-02
Event driven pixi objects and start fetching data

pleek91:
# Description
This is an end of day PR with progress towards a few of objectives.

1. Converts objects to be event driven. 
No more importing `application` into other objects. Simply listen for the `applicationCreated` event which contains the application that can be used. This will help with keeping things independent and not getting stuck in a long order dependent list of functions to call. 

2. Update `scaleX` to be `scaleTime` rather than `scaleLinear`. 
3. Start fetching graph data
This graph data will be used to trigger sizing of the scales and rendering tasks and flow runs

### DIFF
--- a/src/compositions/useRunGraphData.ts
+++ b/src/compositions/useRunGraphData.ts
@@ -1,0 +1,29 @@
+import { UseSubscription, useSubscription } from '@prefecthq/vue-compositions'
+import { ComputedRef, MaybeRefOrGetter, computed, toValue } from 'vue'
+import { RunGraphFetch } from '@/models/RunGraph'
+
+type RunGraphsData = Record<string, UseSubscription<RunGraphFetch>>
+
+function fetchAll(runIds: string[], fetch: RunGraphFetch): RunGraphsData {
+  return runIds.reduce<RunGraphsData>((response, runId) => {
+    response[runId] = useSubscription(fetch, [runId])
+
+    return response
+  }, {})
+}
+
+export type UseRunGraphData = {
+  data: ComputedRef<RunGraphsData>,
+  subscription: UseSubscription<typeof fetchAll>,
+}
+
+export function useRunGraphData(runIds: MaybeRefOrGetter<string[]>, fetch: RunGraphFetch): UseRunGraphData {
+  const parameters = computed<[string[], RunGraphFetch]>(() => [toValue(runIds), fetch])
+  const subscription = useSubscription(fetchAll, parameters)
+  const data = computed(() => subscription.response ?? {})
+
+  return {
+    data,
+    subscription,
+  }
+}
--- a/src/objects/application.ts
+++ b/src/objects/application.ts
@@ -1,15 +1,31 @@
 import { Application } from 'pixi.js'
-import { stage } from '@/objects/stage'
+import { emitter } from '@/objects/events'
 
-export let application: Application
+export let application: Application | null = null
 
-export function createApplication(): void {
+export function startApplication(): void {
+  emitter.on('stageCreated', createApplication)
+}
+
+export function stopApplication(): void {
+  if (!application) {
+    return
+  }
+
+  application.destroy(true, {
+    children: true,
+  })
+}
+
+function createApplication(stage: HTMLDivElement): void {
   application = new Application({
     background: '#1099bb',
     resizeTo: stage,
   })
 
   stage.appendChild(application.view as HTMLCanvasElement)
+
+  emitter.emit('applicationCreated', application)
 
   if (process.env.NODE_ENV === 'development') {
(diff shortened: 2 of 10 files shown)


src/compositions/useRunGraphData.ts
pleek91, line 1: This is a nice outcome of the decision to let implementation provide a rood run id and then a method for fetching graph data for a run id. This composition returns graph data for multiple run ids. So we can start with the root run id and if we encounter a sub flow run we can just add its id to the list of ids we need run data for and this composition will automatically fetch it.

src/objects/box.ts
pleek91, line 1: This is still a dummy object for POC. But you can imaging this box is one task run

src/objects/index.ts
```diff
@@ -1,25 +1,27 @@
...
+  startApplication()
+  startViewport()
+  startScales()
+  startBox()
 
-  createApplication()
-  createViewport()
-  createScales()
-  createBox()
+  startStage(stage)
```
pleek91, line 19: This PR fixed some order dependent code. Now this `startStage` function just must run last. The others can be run in any order. I'd like this to not matter either but for now this is simplest solution. Adding the stage kicks off the rest of the logic.

src/objects/scales.ts
```diff
@@ -1,27 +1,23 @@
...
-}
-
-export function createScaleX(): void {
-  const range: ScaleRange = [0, application.view.width]
-  const domain: ScaleXDomain = [0, 100]
 
-  scaleX = scaleLinear()
+export function startScales(): void {
+  scaleX = scaleTime()
+  scaleY = scaleLinear()
```
pleek91, line 17: scales are originally create without a domain or range. Because the stage doesn't exist yet. Then once the stage is created the initial ranges are set. 

A similar event will be used to set the domain based on the graph data.

src/objects/index.ts
```diff
@@ -1,25 +1,27 @@
...
 }
 
 export function stop(): void {
-  application.destroy(true, {
-    children: true,
-  })
+  emitter.all.clear()
+
+  stopApplication()
+  stopStage()
```
brandonreid, line 26: I wonder if emitting some sort of global stop event might be nice. Like I believe the sprites will need to be destroyed to avoid memory leaks? Things could be different in Pixi 7 though.

brandonreid (approved):
Approving so things can keep rockin. Plenty in flux it seems like but really digging the event emissions stuff happening! Nice that all the things can be more independent and simply listen for their requirements.

src/objects/index.ts
pleek91: We totally could. Though I believe when I'm destroying the app I'm also telling it to destroy the children.

src/compositions/useRunGraphData.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export type UseRunGraphData = {
+  data: ComputedRef<RunGraphsData>,
+  subscription: UseSubscription<typeof fetchAll>,
+}
+
+export function useRunGraphData(runIds: MaybeRefOrGetter<string[]>, fetch: RunGraphFetch): UseRunGraphData {
+  const parameters = computed<[string[], RunGraphFetch]>(() => [toValue(runIds), fetch])
+  const subscription = useSubscription(fetchAll, parameters)
+  const data = computed(() => subscription.response ?? {})
```
znicholasbrown, line 23: This might need to expose options to the consumer to allow polling

src/objects/application.ts
```diff
@@ -1,16 +1,32 @@
...
 
-export let application: Application
+export let application: Application | null = null
 
-export function createApplication(): void {
+export function startApplication(): void {
+  emitter.on('stageCreated', createApplication)
+}
+
+export function stopApplication(): void {
```
znicholasbrown, line 10: Do we not need to detach the emitter here?

znicholasbrown: Nvm I see we do that in the parent caller - but I see this method being called explicitly and not the `stop` method

znicholasbrown (approved):
one question on teardown but not a blocker

src/objects/application.ts
pleek91: `stopApplication` gets called in `stop`. Its my intent that the start* and stop* methods should never get called individually but only by the root start and stop methods. 

Though I could see an argument for having every `emitter.on` in a start method have a corresponding `emitter.off` in the stop method.

I think its probably still a good idea to just clear all the event listeners when the graph is stopped though.

znicholasbrown: Yea probably both are needed - worth noting for now, not worth slowing down any work

PR #223, opened 2023-10-03
Date based axis and better startup logic using "waitFor" methods

pleek91:
# Description
The primary goal of this PR was to introduce the logic for getting and setting the domain for the x axis on the graph. This involves using a subscription to fetch the graph data and a watcher to then set the domain for the axis. That is mostly encapsulated in `useRunGraphDomain.ts`.

While updating the x axis to be date based and respect the domain from the graph data I found some bugs in the initial setup of objects and specifically the scales. This led to a slight refactor, mostly to accomplish 
- Manually stop each object rather than relying on a the vue lifecycle
- Add `waitForEvent` and associated `waitForThing` utilities that make setup dependencies much easier to manage.

### DIFF
--- a/demo/sections/components/RunGraphDemo.vue
+++ b/demo/sections/components/RunGraphDemo.vue
@@ -1,23 +1,39 @@
 <template>
   <p-layout-default class="run-graph-demo">
-    <RunGraph :config="config" :zoom="zoom" class="run-graph-demo__graph" />
-    <p-number-input v-model="zoom" />
+    <RunGraph :config="config" class="run-graph-demo__graph" />
+    <!-- <p-number-input v-model="zoom" /> -->
   </p-layout-default>
 </template>
 
 <!-- eslint-disable camelcase -->
 <script lang="ts" setup>
-  import { endOfHour, startOfHour } from 'date-fns'
-  import { ref } from 'vue'
+  import { randomId } from '@prefecthq/prefect-design'
+  import { endOfHour, startOfHour, startOfDay, endOfDay } from 'date-fns'
+  // import { ref } from 'vue'
   import RunGraph from '@/components/RunGraph.vue'
   import { RunGraphConfig, RunGraphData } from '@/models'
 
-  const zoom = ref(0)
+  // const zoom = ref(0)
   const now = new Date()
 
   const dummy: RunGraphData = {
-    start_time: startOfHour(now),
-    end_time: endOfHour(now),
+    start_time: startOfDay(now),
+    end_time: endOfDay(now),
+    root_node_ids: [],
+    nodes: new Map([
+      [
+        randomId(), {
+          start_time: startOfHour(now),
+          end_time: endOfHour(now),
+          kind: 'task-run',
+          state_name: 'Completed',
+          label: 'bar',
+          child_ids: [],
+          parent_ids: [],
+          id: randomId(),
+        },
+      ],
+    ]),
   }
 
   const config: RunGraphConfig = {
--- a/src/components/RunGraph.vue
+++ b/src/components/RunGraph.vue
@@ -5,6 +5,7 @@
 <script lang="ts" setup>
   import { onUnmounted, onMounted, ref, reactive } from 'vue'
   import { useRunGraphData } from '@/compositions/useRunGraphData'
+  import { useRunGraphDomain } from '@/compositions/useRunGraphDomain'
   import { RunGraphConfig } from '@/models/RunGraph'
   import { start, stop } from '@/objects'
   import { WorkerMessage, worker } from '@/workers/runGraph'
@@ -29,7 +30,9 @@
   }
 
   const runIds = reactive<string[]>([props.config.runId])
-  const { data } = useRunGraphData(runIds, props.config.fetch)
+
+  useRunGraphData(runIds, props.config.fetch)
+  useRunGraphDomain(() => props.config.runId, props.config.fetch)
 
   onMounted(() => {
     if (!stage.value) {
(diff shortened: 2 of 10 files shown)


src/components/RunGraph.vue
```diff
@@ -39,5 +42,7 @@
     start(stage.value)
   })
 
-  onUnmounted(() => stop())
+  onBeforeUnmount(() => {
```
pleek91, line 45: worth calling out that changing this to `onBeforeUnmount` is important because it ensures the correct order of setup -> start -> stop. Using `onUnmounted` meant the `stop` method would get called out of order when vue recycles the component.

src/compositions/useRunGraphDomain.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    return [start, end]
+  })
+
+  watch(domain, domain => {
+    if (!domain) {
+      return
+    }
+
+    setScales({ x: { domain } })
```
pleek91, line 31: Important to point out that this composition is tightly coupled with the graph because its setting the scales for the graph here inside the composition.

demo/sections/components/RunGraphDemo.vue
```diff
@@ -1,23 +1,39 @@
...
   </p-layout-default>
 </template>
 
 <!-- eslint-disable camelcase -->
 <script lang="ts" setup>
-  import { endOfHour, startOfHour } from 'date-fns'
-  import { ref } from 'vue'
+  import { randomId } from '@prefecthq/prefect-design'
+  import { endOfHour, startOfHour, startOfDay, endOfDay } from 'date-fns'
+  // import { ref } from 'vue'
```
znicholasbrown, line 12:
```suggestion
```

demo/sections/components/RunGraphDemo.vue
```diff
@@ -1,23 +1,39 @@
 <template>
   <p-layout-default class="run-graph-demo">
-    <RunGraph :config="config" :zoom="zoom" class="run-graph-demo__graph" />
-    <p-number-input v-model="zoom" />
+    <RunGraph :config="config" class="run-graph-demo__graph" />
+    <!-- <p-number-input v-model="zoom" /> -->
```
znicholasbrown, line 4:
```suggestion
```

demo/sections/components/RunGraphDemo.vue
```diff
@@ -1,23 +1,39 @@
...
-  import { endOfHour, startOfHour } from 'date-fns'
-  import { ref } from 'vue'
+  import { randomId } from '@prefecthq/prefect-design'
+  import { endOfHour, startOfHour, startOfDay, endOfDay } from 'date-fns'
+  // import { ref } from 'vue'
   import RunGraph from '@/components/RunGraph.vue'
   import { RunGraphConfig, RunGraphData } from '@/models'
 
-  const zoom = ref(0)
+  // const zoom = ref(0)
```
znicholasbrown, line 16:
```suggestion
```

demo/sections/components/RunGraphDemo.vue
```diff
@@ -1,23 +1,39 @@
...
+    start_time: startOfDay(now),
+    end_time: endOfDay(now),
+    root_node_ids: [],
+    nodes: new Map([
+      [
+        randomId(), {
+          start_time: startOfHour(now),
+          end_time: endOfHour(now),
+          kind: 'task-run',
+          state_name: 'Completed',
```
znicholasbrown, line 29: I know this is placeholder data but it's probably worth calling out now that we're likely to need state type in this payload as well if we want to support custom states in any capacity in the future... easier to get now than later

src/objects/application.ts
```diff
@@ -32,4 +37,12 @@ function createApplication(stage: HTMLDivElement): void {
...
     (globalThis as any).__PIXI_APP__ = application
   }
+}
+
+export async function waitForApplication(): Promise<Application> {
+  if (application) {
+    return await application
+  }
+
+  return waitForEvent('applicationCreated')
```
znicholasbrown, line 47: It might be worth creating enums for these so we can reference like `event.ApplicationCreated` - depending on how much we plan to use them

src/objects/box.ts
```diff
@@ -2,22 +2,31 @@ import { endOfHour, startOfHour } from 'date-fns'
...
-export function startBox(): void {
-  emitter.on('viewportCreated', createBox)
-  emitter.on('scaleXUpdated', renderBox)
+export async function startBox(): Promise<void> {
+  const viewport = await waitForViewport()
+  createBox(viewport)
+
+  const scales = await waitForScales()
+  renderBox(scales)
+
```
znicholasbrown, line 16: Yeah I think this pattern is turning out nicely

znicholasbrown (approved):

demo/sections/components/RunGraphDemo.vue
pleek91: I'll remove these zoom references in a future PR. For now they're just there so I can test if needed

demo/sections/components/RunGraphDemo.vue
pleek91: Hmm that's a valid point

src/objects/application.ts
pleek91: Personally I think the typesafe strings are preferred. All the same typesafety without the additional import.

src/objects/box.ts
pleek91: Yeah I'm loving this so far

demo/sections/components/RunGraphDemo.vue
pleek91: Will follow up about this.

src/objects/application.ts
znicholasbrown: fair enough 🤷🏻

PR #224, opened 2023-10-04
Convert compositions into objects and add effectScope utilities

pleek91:
# Description
Continuing to refine the object start/stop/wait pattern. Moving the logic for fetching and refreshing the graph domain into an object as well as starting to shape up the rendering of nodes in a similar fashion.


src/objects/domain.ts
```diff
@@ -0,0 +1,63 @@
...
+export async function startDomain(): Promise<void> {
+  const config = await waitForConfig()
+
+  startSubscription(config)
+
+  emitter.on('configUpdated', startSubscription)
+}
+
+export function stopDomain(): void {
+
```
znicholasbrown:
```suggestion
  domain = null
```

src/objects/domain.ts
```diff
@@ -0,0 +1,63 @@
...
+      emitter.emit(event, domain)
+    }, { immediate: true })
+  })
+}
+
+function stopSubscription(): void {
+  scope.stop()
+  subscription = null
+}
```
znicholasbrown, line 55: I'm a little surprised to see this logic here; I'd expect this to be broken out since the subscription necessarily will include data about nodes and anything else the graph is rendering as well. Is the plan that each piece of data will have its own `useSubscription` call + watcher?

znicholasbrown (approved):
thanks for the await changes haha

src/objects/domain.ts
pleek91: I could make the graph data its own object. But then I'd still have to handle sub flow run graph data inline somewhere. So it seemed simpler and more straight forward for objects that need data to fetch it themselves

PR #233, opened 2023-10-09
Add default config values and add real range for y axis

pleek91:
# Description
This PR adds the ability for config values to have default values. So any optional config value will get a default value set for it in `config.ts` and when you get a value from the config it will always have a value. 

A new config value was added for `config.styles.nodeHeight`. This value is used not only to size individual nodes but to also to set the domain for the y axis. 

## Node
At some point the domains for the x axis and possibly the y axis will need to be controlled by the component itself as v-models. This is necessary because at implementation the date range being displayed can be modified externally (by zooming/dragging the events visualization in cloud which is synced with this visualization). Probably going to work on that next so some of this y axis implementation is likely to change. But this seemed like a good stopping point to put up a PR.

### DIFF
--- a/src/models/RunGraph.ts
+++ b/src/models/RunGraph.ts
@@ -1,4 +1,6 @@
 import { ColorSource } from 'pixi.js'
+import { DeepRequired } from 'ts-essentials'
+import { RunGraph } from '..'
 import { StateType } from '@/models/states'
 
 export type RunGraphData = {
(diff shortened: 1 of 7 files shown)


src/models/RunGraph.ts
```diff
@@ -1,4 +1,6 @@
 import { ColorSource } from 'pixi.js'
+import { DeepRequired } from 'ts-essentials'
+import { RunGraph } from '..'
```
znicholasbrown, line 3: This should probably be an aliased import

znicholasbrown (approved):

src/objects/config.ts
```diff
@@ -1,9 +1,32 @@
...
+import { RequiredGraphConfig, RunGraphConfig } from '@/models/RunGraph'
 import { EventKey, emitter, waitForEvent } from '@/objects/events'
 import { waitForScope } from '@/objects/scope'
 
-let config: RunGraphConfig | null = null
+let config: RequiredGraphConfig | null = null
+
+const defaults = {
+  styles: {
+    nodeHeight: 20,
```
brandonreid, line 10: I think previously the height was calculated by the font height + 8px on either side, something like that. A height in the config like this isn't a bad call though. I could see making things more condense with more nodes or something.

src/models/RunGraph.ts
pleek91: Idk why this was here anyway. Its unused.

PR #234, opened 2023-10-09
Domain's vmodel

pleek91:
# Description
Refactors the domains a bit to allow for the domain be a v-model on the component itself.

### DIFF
--- a/src/components/RunGraph.vue
+++ b/src/components/RunGraph.vue
@@ -6,13 +6,23 @@
   import { onBeforeUnmount, onMounted, ref } from 'vue'
   import { RunGraphConfig } from '@/models/RunGraph'
   import { start, stop } from '@/objects'
+  import { RunGraphDomain } from '@/objects/domain'
+  import { emitter } from '@/objects/events'
   // import { WorkerMessage, worker } from '@/workers/runGraph'
 
   const props = defineProps<{
     config: RunGraphConfig,
+    domain?: RunGraphDomain,
+  }>()
+
+  const emit = defineEmits<{
+    (event: 'update:domain', value: RunGraphDomain): void,
   }>()
 
   const stage = ref<HTMLDivElement>()
+
+  emitter.on('domainCreated', domain => emit('update:domain', domain))
+  emitter.on('domainUpdated', domain => emit('update:domain', domain))
 
   // worker.onmessage = onMessage
 
@@ -35,6 +45,7 @@
     start({
       stage: stage.value,
       config: () => props.config,
+      domain: () => props.domain,
     })
   })
 
--- a/src/objects/domain.ts
+++ b/src/objects/domain.ts
@@ -38,12 +38,13 @@ export function stopDomain(): void {
 }
 
 export function setDomain(value: Partial<RunGraphDomain>): void {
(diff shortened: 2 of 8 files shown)


src/components/RunGraph.vue
```diff
@@ -6,14 +6,24 @@
...
+    domain?: RunGraphDomain,
+  }>()
+
+  const emit = defineEmits<{
+    (event: 'update:domain', value: RunGraphDomain): void,
   }>()
 
   const stage = ref<HTMLDivElement>()
 
+  emitter.on('domainCreated', domain => emit('update:domain', domain))
```
znicholasbrown, line 24: Surprised you can't listen to a list of events:
```ts
  emitter.on(['domainCreated', 'domainUpdated'], domain => emit('update:domain', domain))
```

src/objects/domain.ts
```diff
@@ -40,4 +69,50 @@ export async function waitForDomain(): Promise<RunGraphDomain> {
...
+async function watchDomain(domain: () => RunGraphDomain | undefined): Promise<void> {
+  const scope = await waitForScope()
+
+  scope.run(() => {
+    watch(domain, value => {
+      if (value) {
+        setDomain(value)
+      }
+    })
+  })
```
znicholasbrown, line 83: Does this need to be torn down once the scope is disposed?

src/objects/domain.ts
```diff
@@ -1,37 +1,66 @@
...
   domain = null
   stopData()
 }
 
-function getGraphData(runId: string): void {
-  getData(runId, ({ start_time, end_time }) => {
-    const event: EventKey = domain ? 'domainUpdated' : 'domainCreated'
+export function setDomain(value: Partial<RunGraphDomain>): void {
+  if (isEqual(domain, value)) {
+    return
```
znicholasbrown, line 42: I think you might still get unwanted events if the caller passes a partial update that matches the existing domain, like:
```
const domain = { x: 1, y: 2 }

setDomain({ x: 1 })
```
Since `{ x: 1, y: 2} !== { x: 1 }` as this check is comparing

znicholasbrown (approved):

src/components/RunGraph.vue
pleek91: I also felt like that should work. But mitt is purposefully tiny and doesn't have stuff like this. 

If we end up using event busses more I'll probably make our own that implements some of these nice devx features.

src/objects/domain.ts
pleek91: The watcher? That's what the scope does. It's like I mounting a component.

src/objects/domain.ts
pleek91: Ah good catch. I should check x and y independently

src/objects/domain.ts
pleek91: No worries! Scopes are pretty useful for this kind of stuff. The graph has a global scope it creates and automatically stops when the stop method is called. So in any other object like this we can use that scope as a lifecycle

src/objects/domain.ts
pleek91: Updated 👍

PR #235, opened 2023-10-10
Simplify scales and sync viewport date range as a v-model

pleek91:
# Description
Bit of a pivot from preview PRs where the domains for the axes were calculated based on the date range of the run. This removes the domain object altogether in favor of consistently sized scales. This means that layout of runs will be consistent across runs. 

This also pivots on what the date v-model is for implementations to hook into. Previously it was the domain itself. But we don't actually want to change the domain externally. The domain dictates layout. The viewport controls what is visible. So the viewport itself is given a date range and that range is what can be hooked into via the `viewport` v-model. 

## Note
Currently a run will render and might be smaller/larger than the stage. In a future PR the viewport will be scaled to fit the run after the first layout is calculated.


brandonreid (approved):

src/objects/viewport.ts
```diff
@@ -1,21 +1,18 @@
+import isEqual from 'lodash.isequal'
 import { Viewport } from 'pixi-viewport'
-import { Application } from 'pixi.js'
+import { watch } from 'vue'
+import { RunGraphProps } from '..'
```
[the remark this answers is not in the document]
pleek91: ah right, good catch. Really wish the linter would catch those

src/components/RunGraph.vue
```diff
@@ -4,25 +4,22 @@
...
-    config: RunGraphConfig,
-    domain?: RunGraphDomain,
-  }>()
+  // using the props object as a whole
+  // eslint-disable-next-line vue/no-unused-properties
+  const props = defineProps<RunGraphProps>()
 
   const emit = defineEmits<{
-    (event: 'update:domain', value: RunGraphDomain): void,
+    (event: 'update:viewport', value: ScaleXDomain): void,
```
znicholasbrown, line 16: I'm surprised to see `ScaleXDomain` as the emitted type here; the emitter is pass-through emitting the X-scale range just below this, which I think makes more sense given that the viewport is no longer tied to the domain.

This also doesn't take into account the Y scale range, which I know won't be continuous but is probably important given the layout can extend beyond the viewport in both directions

src/objects/scales.ts
```diff
@@ -1,117 +1,112 @@
...
 export type ScaleRange = [start: number, end: number]
 export type ScaleXDomain = [start: Date, end: Date]
 export type ScaleYDomain = [start: number, end: number]
 
-export async function startScales(): Promise<void> {
-  const stage = await waitForStage()
-  setScaleRangesFromStage(stage)
+const { fetch: getData, stop: stopData } = graphDataFactory()
+const DEFAULT_COLUMN_SPAN_SECONDS = 1
+const DEFAULT_COLUMN_SIZE_PIXELS = 20
```
znicholasbrown, line 22: I like this as a control for consistency

src/components/RunGraph.vue
```diff
@@ -4,25 +4,22 @@
...
   }>()
 
   const stage = ref<HTMLDivElement>()
 
-  emitter.on('domainCreated', domain => emit('update:domain', domain))
-  emitter.on('domainUpdated', domain => emit('update:domain', domain))
+  emitter.on('viewportDateRangeUpdated', range => emit('update:viewport', range))
```
znicholasbrown, line 21: Does this not get emitted as soon as the viewport finishes loading? I'm not seeing the associated v-model populated until the viewport changes

znicholasbrown (commented):
A couple of non-blocking comments; I generally agree with the motivation and like how much simpler the scales have gotten 👍🏻

src/components/RunGraph.vue
pleek91: `ScaleXDomain` is just a date tuple. Happens to be the same type as the viewport visible dates. Should I make a separate type? 

Hadn't considered the y axis since our current implementation only syncs the x axis. Probably could expose that, however might get weird because the viewport is a specific size (set by the canvas element). So you have to maintain that aspect ratio. So if you changed the dates you'd have to change the y axes to match. Which would get weird IMO.

src/components/RunGraph.vue
pleek91: you might be right, will investigate

pleek91: Fixed in https://github.com/PrefectHQ/graphs/pull/236

src/components/RunGraph.vue
znicholasbrown: Yeah I guess I'd expect the viewport v-model to reflect the entirety of the viewport and not one dimension of it, though I recognize it feels like the language should be the same; I don't think there's an immediate need to address this just thinking through camera controls later down the line, they could get pretty wonky or be very limited in this implementation

pleek91: yeah makes sense. Just not something we need right now I don't think.

PR #236, opened 2023-10-10
Add the ability to center the viewport to the world

pleek91:
# Description
Adds a `centerViewport` method that will automatically scale and center the graph in the viewport

Also fixes an issue where the viewportDateRange was not set when the viewport was started.

### DIFF
--- a/src/objects/nodes.ts
+++ b/src/objects/nodes.ts
@@ -97,6 +97,7 @@ async function renderNode(node: RunGraphNode): Promise<void> {
 }
 
 async function createNode(node: RunGraphNode): Promise<NodeSprites> {
+  const container = await waitForContainer()
   const { inter } = await waitForFonts()
   const graphics = new Graphics()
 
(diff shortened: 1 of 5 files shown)


src/objects/nodes.ts
```diff
@@ -78,8 +109,8 @@ async function createNode(node: RunGraphNode): Promise<NodeSprites> {
     label,
   })
 
-  viewport.addChild(graphics)
-  viewport.addChild(label)
+  container!.addChild(graphics)
+  container!.addChild(label)
```
znicholasbrown, line 113: why the non-null assertion? is this coming from pixi and not well-typed?

znicholasbrown (approved):
one question otherwise lgtm

also where'd my deploy preview go

src/objects/nodes.ts
pleek91: oh I forgot to come back to this. I didn't care to fix it until I knew it was working. Will update

pleek91: Fixed!

pleek91:
Deploy preview probably didn't show up until this was updated to PR into main, this was chained off another branch originally.

PR #237, opened 2023-10-11
Draw nodes prior to doing layout

pleek91:
# Description
This is a bit of a pivot prior to implementing the actual layout worker. In the v1 we first send the nodes to the layout worker, calculate the layout, then we render. This means that we don't actually know the size of individual nodes because we don't know if the label fits inside the box or not. So we calculate the do the box and label calculations inside the worker. This is a little problematic because we don't actually know the width of the label at this point. So we calculate it based on the label's length multiplied by a approximate character width. All this is done in the worker when calculating the layout. 

What this PR starts implementing is breaking apart the rendering of nodes and the positioning of nodes. So rather than
- Get nodes data
- Calculate layout (in worker)
- Render, position, and add nodes to the canvas

We'll end up with
- Get nodes data
- Render nodes (but don't add them to the canvas yet!)
- Calculate layout (in worker, based on actual size of nodes)
- position and add nodes to the canvas

I believe this will simplify the layout complexity. It becomes "figure out where the nodes go" rather than "figure out the size of the nodes and figure out where they go".

I also think this means the layouts can be determined independent of the type of scales. This would mean that we could possibly reuse the "nearest parent" layout when rendering a run in a DAG like layout. The layout can just care about x,y and not care about dates. 

## Note about `nodeRenderKey`s
I also added the concept of "render keys" in this PR. Essentially, when rendering a node we can check if the rendering we already have is what we're going to end up with. So we can skip rendering and reuse what has already been rendered. If the key does not match, the existing pixi objects are updated rather than being recreated. I believe this will increase performance when viewing graphs where the root node (flow run, task run) is "running".

## Deployment preview missing layout
While pivoting I removed the temporary waterfall layout that was built into the nodes object. That will get re-added in the next PR in the layout worker.

### DIFF
--- a/src/components/RunGraph.vue
+++ b/src/components/RunGraph.vue
@@ -1,5 +1,5 @@
 <template>
-  <div ref="stage" />
+  <div ref="stage" class="run-graph" />
 </template>
 
 <script lang="ts" setup>
@@ -49,3 +49,10 @@
     stop()
   })
 </script>
+
+<style>
+.run-graph > canvas {
+  width: 100%;
+  height: 100%;
+}
+</style>
--- a/src/objects/application.ts
+++ b/src/objects/application.ts
@@ -27,6 +27,7 @@ function createApplication(stage: HTMLDivElement): void {
     backgroundAlpha: 0,
     resizeTo: stage,
     antialias: true,
+    resolution: window.devicePixelRatio,
   })
 
   stage.appendChild(application.view as HTMLCanvasElement)
--- a/src/objects/config.ts
+++ b/src/objects/config.ts
@@ -7,7 +7,7 @@ let config: RequiredGraphConfig | null = null
 
 const defaults = {
   animationDuration: 200,
-  nodeRenderKey: (node) => `${node.kind},${node.start_time},${node.end_time},${node.state_type},${node.label}`,
+  nodeRenderKey: (node) => `${node.id},${node.kind},${node.start_time},${node.end_time},${node.state_type},${node.label}`,
   styles: {
     nodeHeight: 30,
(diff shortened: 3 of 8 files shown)


src/components/RunGraph.vue
pleek91, line 1: Unrelated, just makes sure the canvas is responsive to the stage.

src/objects/application.ts
```diff
@@ -27,6 +27,7 @@ function createApplication(stage: HTMLDivElement): void {
     backgroundAlpha: 0,
     resizeTo: stage,
     antialias: true,
+    resolution: window.devicePixelRatio,
```
pleek91, line 30: Unrelated, just fixes some blurriness.

src/objects/config.ts
```diff
@@ -7,8 +7,10 @@ let config: RequiredGraphConfig | null = null
 
 const defaults = {
   animationDuration: 200,
+  nodeRenderKey: (node) => `${node.id},${node.kind},${node.start_time},${node.end_time},${node.state_type},${node.label}`,
```
pleek91, line 10: I thought about using a hashing function and just hashing the node itself. But I feel like this is probably much faster even if its more brittle. You have to make sure to add any properties here that effect the rendered node.

pleek91: Could probably improve this quite a bit by having this just return an array of keys or something. For now this is good enough.

brandonreid (approved):

znicholasbrown (approved):
thanks for the thorough description - no concerns here, good to know you can render ahead of time, i think that reasoning is sound

PR #238, opened 2023-10-12
Calculate layout in the worker

pleek91:
# Description
First layout implementation in the worker. A simple waterfall layout. But the important piece is the work is sent a `GraphPreLayout` and receives back a `GraphPostLayout`. The worker is given the basic information it needs to calculate the y axis and sends back the the layout with the correct y value for each node.

### DIFF
--- a/src/objects/nodes.ts
+++ b/src/objects/nodes.ts
@@ -6,6 +6,7 @@ import { emitter } from '@/objects/events'
 import { waitForFonts } from '@/objects/fonts'
 import { waitForNodesContainer } from '@/objects/nodesContainer'
 import { waitForScales } from '@/objects/scales'
+import { centerViewport } from '@/objects/viewport'
 import { exhaustive } from '@/utilities/exhaustive'
 import { graphDataFactory } from '@/utilities/graphDataFactory'
 import { WorkerLayoutMessage, WorkerMessage, getLayoutWorker } from '@/workers/runGraph'
@@ -36,7 +37,9 @@ function onMessage({ data }: MessageEvent<WorkerMessage>): void {
   }
 }
 
-function handleLayoutMessage({ layout }: WorkerLayoutMessage): void {
+async function handleLayoutMessage({ layout }: WorkerLayoutMessage): Promise<void> {
+  const config = await waitForConfig()
+
   layout.forEach((layout, nodeId) => {
     const objects = graphObjects.get(nodeId)
 
@@ -47,9 +50,11 @@ function handleLayoutMessage({ layout }: WorkerLayoutMessage): void {
 
     const { x, y } = layout
 
-    objects.container.position = { x, y }
+    objects.container.position = { x, y: config.styles.nodeHeight * y }
     objects.container.visible = true
   })
+
+  centerViewport()
 }
 
 export async function startNodes(): Promise<void> {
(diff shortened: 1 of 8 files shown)


src/objects/nodes.ts
```diff
@@ -117,6 +149,7 @@ function createContainer(): Container {
   const container = new Container()
 
   container.eventMode = 'none'
+  container.visible = false
```
pleek91, line 152: Noteworthy that rather than waiting to add the node container objects to pixi we're adding them but setting their visibility to `false`. Once the layout comes back from the worker and all the node positions are updated then their visibility is set to `true`.

src/utilities/graphDataFactory.ts
```diff
@@ -27,7 +27,7 @@ export function graphDataFactory(): GraphDataFactory {
 
       watch(() => subscription.response, response => {
         if (response) {
-          callback(response)
+          callback(toRaw(response))
```
pleek91, line 30: This just makes sure we don't pass any reactive data. Removes vue's proxies which makes the data handling a little cleaner and lighter.

src/workers/runGraph.ts
```diff
@@ -1,20 +1,30 @@
...
 
-export interface IRunGraphWorker extends Omit<Worker, 'postMessage'> {
-  postMessage: (command: ClientMessage, transfer: Transferable[]) => void,
+export interface IRunGraphWorker extends Omit<Worker, 'postMessage' | 'onmessage'> {
+  postMessage: (command: ClientMessage, transfer?: Transferable[]) => void,
   onmessage: ((this: Worker, ev: MessageEvent<WorkerMessage>) => void) | null,
 }
 
-export const worker: IRunGraphWorker = new RunGraphWorker()
+export function getLayoutWorker(onmessage: IRunGraphWorker['onmessage']): IRunGraphWorker {
```
pleek91, line 24: Little abstraction so we can have multiple workers. 1:1 for each run. So when we get to sub flows we don't have to worry about overwhelming a single worker.

src/objects/nodes.ts
```diff
@@ -139,6 +172,7 @@ async function updateNodeBox(node: RunGraphNode, box: Graphics): Promise<Graphic
   const boxWidth = boxRight - boxLeft
   const boxHeight = config.styles.nodeHeight - config.styles.nodeMargin * 2
 
+  box.lineStyle(1, 0x0, 1, 2)
```
pleek91, line 175: Removing this in https://github.com/PrefectHQ/graphs/commit/aa7f73b50e41c69ccad446a4eb6ae3ab9508989d somehow slows everything down. Box rendering will likely change and use sprites but adding it back for now just to keep moving. No idea why removing this extra line around the boxes makes it excruciatingly slow

brandonreid (approved):
I'm sure some of this will continue to evolve once you get to updating data and stuff 🚂

pleek91:
> I'm sure some of this will continue to evolve once you get to updating data and stuff 🚂

Right, everything here is subject to change. Painting with a paint roller at this point lol. We'll get to the pin stripes!

src/workers/runGraph.ts
znicholasbrown: I like this abstraction a lot, I might steal this for use elsewhere

src/utilities/exhaustive.ts
```diff
@@ -0,0 +1,3 @@
+export function exhaustive(value: never): void {
```
znicholasbrown, line 1: ❤️

src/objects/nodes.ts
znicholasbrown: ??

znicholasbrown (approved):
no concerns here other than that line style slowing things down

PR #240, opened 2023-10-12
Add center and fullscreen actions

pleek91:
# Description
Adds the center and fullscreen actions directly to the RunGraph component. Previously these were implementation details in ui-library. But I think it makes sense to have these built in. 

The settings icon to switch layouts will get added later. I also didn't wire up the keyboard shortcuts yet

image

### DIFF
--- a/src/components/RunGraph.vue
+++ b/src/components/RunGraph.vue
@@ -1,22 +1,48 @@
 <template>
-  <div ref="stage" class="run-graph" />
+  <div class="run-graph" :class="classes.root">
+    <div ref="stage" class="run-graph__stage" />
+    <div class="run-graph__actions">
+      <p-button title="Recenter Timeline" icon="Target" flat @click="() => centerViewport({ animate: true })" />
+      <p-button title="View Timeline in Fullscreen" icon="ArrowsPointingOutIcon" flat @click="toggleFullscreen" />
+    </div>
+  </div>
 </template>
 
 <script lang="ts" setup>
-  import { onBeforeUnmount, onMounted, ref } from 'vue'
+  import { computed, onBeforeUnmount, onMounted, ref } from 'vue'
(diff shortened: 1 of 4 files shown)


src/components/RunGraph.vue
```diff
@@ -1,23 +1,49 @@
...
 <script lang="ts" setup>
-  import { onBeforeUnmount, onMounted, ref } from 'vue'
+  import { computed, onBeforeUnmount, onMounted, ref } from 'vue'
   import { RunGraphProps } from '@/models/RunGraph'
-  import { ScaleXDomain, start, stop } from '@/objects'
+  import { ScaleXDomain, start, stop, centerViewport } from '@/objects'
   import { emitter } from '@/objects/events'
 
   // using the props object as a whole
   // eslint-disable-next-line vue/no-unused-properties
```
znicholasbrown, line 18: this comment can probably be deleted now

znicholasbrown (commented):
Screenshot 2023-10-12 at 6 31 01 PM

In fullscreen mode the aspect ratio of nodes seems to get skewed incorrectly - you might be missing a call to `resizeViewport`? I'm not totally sure when `stageUpdated` gets called

znicholasbrown (approved):

pleek91:
I also had that thought but couldn't decide for sure the longer I looked at it. I'll figure out what's going on there 👍

src/components/RunGraph.vue
pleek91: Nah eslint still complains about the props being unused. These are the reasons I don't like the compiler time macros. Their not real js/ts so they have to write all these special cases and handling into the ecosystem. And seems like the vue/no-unused-properties just isn't smart enough to tell I'm using the props const so all the props are used.

PR #246, opened 2023-10-16
Convert node objects to classes

pleek91:
# Description
Pivoting some of the global objects into individual services to support nesting of nodes (for sub flow runs and hopefully groups at some point). 

Most of this is just moving things around but there are some new concepts. I've tried to comment on each file the important changes or call out new concepts.


src/objects/config.ts
```diff
@@ -8,7 +8,6 @@ let config: RequiredGraphConfig | null = null
 
 const defaults: Omit<RequiredGraphConfig, 'runId' | 'fetch'> = {
   animationDuration: 500,
-  nodeRenderKey: (node) => `${node.id},${node.kind},${node.start_time},${node.end_time},${node.state_type},${node.label}`,
```
pleek91, line 11: Remove this config in favor if a more dynamic version in nodeContainerService that should work no matter what keys are used.

src/models/layout.ts
pleek91, line 1: Starting to bake in different "layout modes". Currently hard coded to `time` and `waterfall`. 

- Horizontal: 'time`, `dag`
- Vertical: `waterfall`, `nearest-parent`

src/objects/index.ts
```diff
@@ -29,7 +28,6 @@ export function start({ stage, props }: StartParameters): void {
   startFonts()
   startStage(stage)
   startConfig(props)
-  startNodesContainer()
```
pleek91, line 32: There's no longer a global object for the root nodes container. It can be retrieved from the viewport using `viewport.getChildByName(DEFAULT_NODES_CONTAINER_NAME)`.

src/objects/layout.ts
pleek91, line 1: This is the start of a global object for layout modes. There will be events around the layout changing and the gear menu on the graph will be the input for this.

src/objects/nodes.ts
pleek91, line 1: Now this is just in charge of initializing the first NodeContainerService (which does layout and rendering) rather than being responsible for the layout and rendering of the nodes themselves.

src/objects/scales.ts
pleek91, line 1: We still need some global scales for the visible viewport dates. So reusing the NodePositionService here rather than having dedicated scales.

src/objects/viewport.ts
```diff
@@ -138,11 +143,14 @@ async function startViewportDateRange(): Promise<void> {
...
 async function updateViewportDateRange(): Promise<void> {
   const viewport = await waitForViewport()
-  const { scaleX } = await waitForScales()
-  const left = scaleX.invert(viewport.left)
-  const right = scaleX.invert(viewport.right)
+  const scales = await waitForScales()
+  const left = scales.getPositionFromXPixels(viewport.left)
+  const right = scales.getPositionFromXPixels(viewport.right)
+
+  if (left instanceof Date && right instanceof Date) {
```
pleek91, line 150: Starting to see the different modes show up in types. In `dag` mode the x scale is not date based. So checking here if the left/right values are dates before emitting them up.

src/services/nodeContainerService.ts
pleek91, line 1: This service is responsible for rendering a node and returning its "pre layout" value. The individual pieces will probably be abstracted into their own services to support things like sub flow runs and nodes in a "running" state

src/services/nodeOffsetService.ts
pleek91, line 1: Converted the `axisBumpFactory` into this `NodeOffsetService`. Almost identical functionality except the `bump` method from `axisBumpFactory` is now named `getTotalOffset`.

src/services/nodePositionService.ts
pleek91, line 1: Here you can really see the separate layout modes. Each mode has a potentially different type of scale with different settings. 

The primary purpose of this service is to convert Pixels to Positions and vice-versa. 

- Pixels - literal x,y coordinates on the graph (number, number)
- Positions - scale x,y coordinates (date, row)

src/services/nodePositionService.ts
```diff
@@ -0,0 +1,145 @@
...
+      const endOffset = this.offsets.getOffset(index)
+      const end = start + rowHeight + endOffset
+
+      return [start, end]
+    }).flat()
+
+    const axis = scaleLinear().domain([0, maxValue]).range(range)
+
+    return axis.invert(value)
+  }
```
pleek91, line 107: This is a work in progress but will be finished once we get to interactions

src/services/nodesContainerService.ts
pleek91, line 1: This service is responsible for fetching and refreshing run data as well as rendering nodes in the correct position via a layout worker.

src/services/nodeContainerService.ts
pleek91, line 1: This service is responsible for rendering a node and returning its "pre layout" information to the NodesContainerService.

Individual pieces will likely be abstracted to support sub flow runs and nodes that are "running"

src/utilities/promiseFactory.ts
pleek91, line 1: Little utility for "externally resolvable" promises. Used in the NodeContainerService to create a promise that can be awaited until the node is finished rendering.

brandonreid (approved):
I gave this a scan and approving to keep things trucken. I'll try and take a deeper look in the morning though.

PR #249, opened 2023-10-17
Abstract rendering logic for each `kind` of node

pleek91:
# Description
Adds the ability for nodes to be rendered differently based on the `kind`. There are currently two kinds: `task-run` and `flow-run`. In this PR they are rendered almost identically. Just an extra `FLOW RUN` added to the label. But this gives a nice separate point to start building out flow runs on the graph. 

image


src/utilities/getLabelPositionRelativeToBox.ts
pleek91, line 1: Removing this utility in favor of each node service having its own version. Since a flow run needs to make room for the little drop down button the logic will be slightly different for both.

demo/sections/components/RunGraphDemo.vue
pleek91, line 1: Just simplifying this a bit so its all included in the reviver.

src/services/containerService.ts
pleek91, line 1: I'm not 100% on this. I could just extend the `Container` class. But my thought so far has been to have a service that has a pixi object as a property rather than extending the pixi classes. That way we don't have naming collisions or accidentally override something from the base class.

src/services/nodeBoxService.ts
pleek91, line 1: Renamed all instances of `position` to `positionService` to be more clear and not conflict with `position` with is the `Pixel` value of an object.

src/services/nodeContainerService.ts
pleek91, line 1: This service is now only responsible for the render cycles and layout calculations. The actual rendering happens in the individual node services.

brandonreid (approved):
Approving to keep the train going. No strong opinion on the containerService (first impression is that it's maybe a bit superfluous.

Keep going dude!! 🚀

PR #251, opened 2023-10-18
Refactor services into factories and replace mitt with custom events factory

pleek91:
# Description

## Factories
Almost every time I build something using classes and inheritance I regret it. Quickly ended up there with the node and position services I had. Main issue is up until those services basically everything was async. But service constructors cannot be async. Which means the types get really weird and annoying when you need one of your properties to be something async. Replacing services with factories means we can `await` wherever and whenever we want.

## Events
`mitt` is a great little library. But I kept wanting a `once` method which it lacks. Rather than trying to enhance mitt I decided to just roll my own which is probably simpler than mitt and removes a dependency. 

## Layout
I had the concept of a "pre layout" where only the necessary information was sent to the layout worker. And the layout worker didn't care about dates or relationships. Came to the realization that wasn't going to work and the complexity of a "pre" and "post" worker layout was causing issues. Simplified this by making a "NodeLayoutRequest" which contains
- layout mode settings
- nodes and their actual widths

Then a `NodeLayoutResponse` is returned from the worker that just has `Map<nodeId, { x: number, y: number}>` which can be easily used to position the nodes.


src/factories/box.ts
```diff
@@ -0,0 +1,34 @@
+import { differenceInMilliseconds, millisecondsInSecond } from 'date-fns'
+import { Graphics } from 'pixi.js'
+import { DEFAULT_TIME_COLUMN_SIZE_PIXELS } from '@/consts'
+import { RunGraphNode } from '@/models/RunGraph'
+import { waitForConfig } from '@/objects/config'
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export async function nodeBoxFactory() {
```
pleek91, line 8: A lot of these factories infer their return type. Doing this for speed now to avoid maintaining a bunch of types for functions. Planning on coming back through and resolving these linting ignores once things are stable.

src/factories/box.ts
```diff
@@ -0,0 +1,34 @@
...
+  const box = new Graphics()
+
+  async function render(node: RunGraphNode): Promise<Graphics> {
+    const { background } = config.styles.node(node)
+
+    const right = node.start_time
+    const left = node.end_time ?? new Date()
+    const seconds = differenceInMilliseconds(left, right) / millisecondsInSecond
+    const boxWidth = seconds * DEFAULT_TIME_COLUMN_SIZE_PIXELS
+    const boxHeight = config.styles.nodeHeight - config.styles.nodeMargin * 2
```
pleek91, line 19: Worth calling out, this calculation used to depend on the x axis scale. But I realized last night it doesn't need to since we know how wide each second should be on the graph. Simplifies the dependencies here so it doesn't.

src/factories/offsets.ts
pleek91, line 1: This file is exactly the same. Was a factory, then a class, back to being a factory 🤪

src/factories/position.ts
pleek91, line 1: Super simplified this. Just creating a scale for the x axis based on the mode and some static values

src/factories/settings.ts
pleek91, line 1: At some point this will use some more dynamic values for the `timeSpanPixels` and `dagColumnSize` so the x axis can be scaled independently.

src/objects/nodes.ts
```diff
@@ -1,19 +1,46 @@
...
-    runId: config.runId,
-    parent: viewport,
-  })
+  nodes = await nodesContainerFactory(config.runId)
+
+  viewport.addChild(nodes.container)
+
+  nodes.container.alpha = 0
+
+  nodes.events.once('rendered', center)
```
pleek91, line 19: Yay `once`!

src/objects/scale.ts
pleek91, line 1: This got renamed. Most of this is not new. Just simplified.

brandonreid (approved):
dropping hammers

PR #252, opened 2023-10-18
Render sub flow runs

pleek91:
# Description
Clicking on a flow run on the graph will the flow run on the graph. Bumping other content out of the way. 

For now it just renders right on to of the flow run that was clicked on and you cannot collapse it. 

WIP PR to try and keep things a little smaller. 

The last node in the demo is a flow run. Clicking on it will render another demo flow on the graph. 

image

### DIFF
--- a/src/factories/flowRun.ts
+++ b/src/factories/flowRun.ts
@@ -6,6 +6,8 @@ import { Pixels } from '@/models/layout'
 import { RunGraphNode } from '@/models/RunGraph'
 import { waitForConfig } from '@/objects/config'
 
+export type FlowRunContainer = Awaited<ReturnType<typeof flowRunContainerFactory>>
+
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
 export async function flowRunContainerFactory() {
   const container = new Container()
@@ -50,6 +52,7 @@ export async function flowRunContainerFactory() {
   }
 
   return {
+    kind: 'flow-run' as const,
     container,
     render,
   }
--- a/src/factories/flowRun.ts
+++ b/src/factories/flowRun.ts
@@ -10,21 +10,24 @@ import { waitForConfig } from '@/objects/config'
 export type FlowRunContainer = Awaited<ReturnType<typeof flowRunContainerFactory>>
 
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
-export async function flowRunContainerFactory() {
+export async function flowRunContainerFactory(node: RunGraphNode) {
   const container = new Container()
   const { label, render: renderLabel } = await nodeLabelFactory()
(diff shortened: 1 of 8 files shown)


src/factories/flowRun.ts
```diff
@@ -28,6 +35,23 @@ export async function flowRunContainerFactory() {
...
+
+    container.emit('resized')
+  }
+
+  async function drawNodes(): Promise<void> {
+    const nodes = await nodesContainerFactory('foo')
+    container.addChild(nodes.container)
+
+    nodes.container.on('rendered', () => container.emit('resized'))
+  }
```
pleek91, line 53: This will evolve quite a bit. Crud implementation just as a POC of the various pieces being used here.

src/factories/flowRun.ts
```diff
@@ -50,6 +74,7 @@ export async function flowRunContainerFactory() {
   }
 
   return {
+    kind: 'flow-run' as const,
```
pleek91, line 77: Added `kind` as a a discriminating union for the `nodeContainerFactory`

src/factories/nodes.ts
```diff
@@ -13,18 +12,13 @@ import { WorkerLayoutMessage, WorkerMessage, layoutWorkerFactory } from '@/worke
 
 export type NodesContainer = Awaited<ReturnType<typeof nodesContainerFactory>>
 
-type Events = {
-  rendered: void,
-}
```
pleek91, line 18: I removed the factory specific events in favor of augmenting the built in pixi events. That way I can emit directly on containers which is simpler and avoids some event bubbling.

src/factories/offsets.ts
```diff
@@ -12,24 +11,24 @@ type RemoveOffsetParameters = {
...
   function getOffset(axis: number): number {
-    const values = offsets.get(axis)
-
-    if (!values) {
-      return 0
-    }
+    const values = offsets.get(axis) ?? []
 
-    return Math.max(...values.values(), 0)
+    return Math.max(...values.values(), config.styles.nodeHeight)
```
pleek91, line 25: This is evolving a bit to factor in the entire row height and not just the "offset" the name of this utility is likely to change to reflect but waiting to see how this shakes out

src/factories/flowRun.ts
```diff
@@ -28,6 +38,14 @@ export async function flowRunContainerFactory() {
     return container
   }
 
+  function toggle(): void {
+    open = !open
+
+    if (open) {
+      renderNodes()
+    }
+  }
```
pleek91, line 47: This will evolve quite a bit. Just a POC

zhen0 (approved):
Thanks Craig!

PR #253, opened 2023-10-19
Allow user to set the view to trace or dependency

pleek91:
# Description
Adds the graph settings to the graph actions. This currently has two settings. Only view is supported at this point.
- View - How nodes are rendered along the x axis
    - Dependency - All nodes are the same size and ordered by how far from the "root" of the graph they are
    - Trace - Nodes are rendered according to run time
- Layout - How nodes are rendered along the y axis
    - Waterfall - Fastest layout which renders nodes along the y axis in order of when they occurred (first nodes at the top, last at the bottom)
    - Nearest Parent -  Slower layout which aims to render nodes more logically based on where a nodes parent nodes are located.
   
image
image

### DIFF
--- a/src/components/RunGraphSettings.vue
+++ b/src/components/RunGraphSettings.vue
@@ -1,0 +1,100 @@
+<template>
+  <p-pop-over
+    class="run-graph-settings"
+    auto-close
+    :placement="placement"
+  >
+    <template #target="{ toggle }">
+      <p-button
+        aria-label="Run Graph Options"
+        icon="CogIcon"
+        flat
+        @click="toggle"
+      />
+    </template>
+
+    <p-overflow-menu class="run-graph-settings__menu">
+      <p-label label="View">
+        <p-radio-group v-model="horizontal" :options="horizontalOptions">
+          <template #label="{ option }">
+            {{ option.label }}
+          </template>
+        </p-radio-group>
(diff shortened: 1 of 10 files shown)


src/components/RunGraphSettings.vue
```diff
@@ -0,0 +1,100 @@
...
+      <p-button
+        aria-label="Run Graph Options"
+        icon="CogIcon"
+        flat
+        @click="toggle"
+      />
+    </template>
+
+    <p-overflow-menu class="run-graph-settings__menu">
+      <p-label label="View">
```
brandonreid, line 17: Nbd for now, but I do feel this distinction is tough to groc. The x access can be time bound or like.. "linear"? then the y axis or sorting can be arranged differently. Probably just semantics, maybe a simple flat list of layouts is better, not totally sure off the cuff.

brandonreid (approved):

src/components/RunGraphSettings.vue
pleek91: Yeah idk here. Technically the "view" is how the x axis is rendered and the "layout" is how the y axis is rendered. And it makes sense to mix and match both. But I'm also not 100% on this UX for controlling it. Important part is we can control it haha.

PR #255, opened 2023-10-20
Layout organization and optimization of getColumns utility

pleek91:
# Description
Organizes the layout code to have separate files and methods for horizontal and vertical layout. Also adds the vertical layout settings which get sent to the worker when doing layout. 

## `getColumns`
There are currently two methods for getting columns for the dependency view. One is significantly faster, and runs successfully on the largest and most connected data set we currently have to test with. But @chrisguidry and myself are not 100% confident it will run successfully for flows of any size without getting a `Maximum call stack exceeded` error. We were getting that error originally and optimized the code to where we are no longer erroring. But did we fix it for every flow? Or would a larger and more connected flow also exceed the stack. We're not sure. 

The good news is when we were getting errors it was erroring pretty quickly. So I've added a try/catch that if the faster method fails it will try again with the slower method that we are more confident will work for flows of any size.


src/utilities/columns.ts
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * This function processes the children of each node in the `nodeIds` array and calculates their columns based on the columns of their parents.
+ *
+ * @param nodeIds - The IDs of the nodes whose children should be processed.
+ * @param nodes - The nodes in the run graph.
+ * @param columns - A `Map` object that maps each node ID to its column.
+ * @returns A `Map` object that maps each node ID to its column.
+ * @throws An error if a node ID in `nodeIds` is not found in `nodes`.
+ */
+function getChildrenColumns(nodeIds: string[], nodes: RunGraphNodes, columns: NodeColumns): NodeColumns {
```
brandonreid, line 64: Could this be an async function to avoid the stack overflow?

pleek91: You said this earlier and its just clicking with me what you mean. Let me test some things

pleek91: Discussed this via slack. But making this async cuts its performance over half. Still faster than the "getColumnsSlower" method though. 

Might optimize this further but for now I want to move on to other pieces.

brandonreid (approved):

PR #257, opened 2023-10-22
Use textures and sprites for node bars

pleek91:
# Description
Replacing all the node bar graphics with sprites and textures which are much more performant.


src/factories/bar.ts
pleek91, line 1: Generic bar factory with the ability to disable the "caps"

src/factories/nodeBar.ts
pleek91, line 1: nodeBarFactory is a node specific implementation of barFactory

src/factories/rectangle.ts
pleek91, line 1: Creates a single pixel texture

src/factories/cap.ts
```diff
@@ -0,0 +1,58 @@
...
+  graphic.beginFill('#fff')
+  graphic.drawRoundedRect(0, 0, radius * 2, height, radius)
+  graphic.endFill()
+
+  const cap = application.renderer.generateTexture(graphic, {
+    // drew a rounded rectangle and then just using half of the graphic to get just the left "cap"
+    region: new Rectangle(0, 0, radius, height),
+
+    // manually bumping up the resolution to keep the border radius from being blurry
+    resolution: 10,
```
brandonreid, line 23: This is a little sus to me, like the global resolution should make it so that things render without dithering unless zoomed in to an unusable degree (the live version clamps the zoom in). I believe this also makes your texture a lot larger, but probably not big deal.

src/factories/nodeBar.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  function getTotalWidth(node: RunGraphNode, borderRadius: number): number {
+    if (layout.horizontal === 'trace') {
+      const right = node.start_time
+      const left = node.end_time ?? new Date()
+      const seconds = differenceInMilliseconds(left, right) / millisecondsInSecond
+      const width = seconds * DEFAULT_TIME_COLUMN_SIZE_PIXELS
+
+      // this means the min node size is 18px. Is that correct?
+      return Math.max(width, borderRadius * 2)
```
brandonreid, line 39: I believe `borderRadius` is set to 8px, so the min width would be 16px. Glad you caught the min-width there, key for some level of visibility / clickability.

src/factories/rectangle.ts
```diff
@@ -0,0 +1,22 @@
+import { useSubscription } from '@prefecthq/vue-compositions'
+import { Graphics, RenderTexture, Sprite } from 'pixi.js'
+import { waitForApplication } from '@/objects'
+
+async function rectangle(): Promise<RenderTexture> {
+  const application = await waitForApplication()
+
+  const rectangle = new Graphics()
+  rectangle.beginFill('#fff')
+  rectangle.drawRect(0, 0, 1, 1)
```
brandonreid, line 10: Since this creates a bitmap, then stretches it wide/tall enough to cover the geometry it's applied to, 1px might be a bit aggressive? You can imagine if it needs to stretch over a 500px wide thing, it'll need to scale to something well into the 10s of thousands to cover it, whereas a 10px x 10px even gives more solid room to chew on? I don't know how much it theoretically matters though.

brandonreid (approved):

src/factories/nodeBar.ts
pleek91: Right it's 2x the border radius. Not 18px. That tracts with what you'd expect?

brandonreid: Yeah totally

src/factories/rectangle.ts
pleek91: Since it has hard edges I don't think it matters? But I have no preference here. Just seemed like the most flexible and smallest texture. Tbh I kinda feel like we shouldn't need this but didn't want to stray to far from an implementation we know works.

src/factories/cap.ts
pleek91: Correct it creates a larger texture. 300px tall for a 30px cap. But I found even using the default resolution it immediately got blurry when zooming. I believe because it's creating the texture at the exact dimensions requested. So any rounded edges are going to pixilate as soon as you zoom even the tiniest bit. Increasing the resolution means packing in more pixels and gets much cleaner edges. We can play with just how high of a resolution is needed. But I believe it must be higher than the default in order to zoom at all.

src/factories/rectangle.ts
brandonreid: ah you might be right on the square.. At some point in the past I was able to see the texture applied to the geometry in the pixi dev tools, but can't seem to figure out how I did that. Seems to be working just fine as is.

brandonreid: Ah you might be right on the square thing.. At some point I was able to see the texture scaled in the pixi dev tools, but not able to now. Looks great locally so nbd

PR #258, opened 2023-10-23
Nearest parent layout

brandonreid:
This is largely a copy over of the existing nearestParent layout code. Couple updates:

- It leverages the widths and x positions provided by the horizontal layout.
- It's been updated so it doesn't need to create unnecessary copies of stores.

Still lots of optimization and re-thinking that can be done here but it is generally working for now. The example used still does have an overlapping node that will need to be investigated though. Will continue to work on this.

### DIFF
--- a/src/workers/layouts/nearestParentVertical.ts
+++ b/src/workers/layouts/nearestParentVertical.ts
@@ -1,0 +1,390 @@
+import { HorizontalLayout } from '@/workers/layouts/horizontal'
+import { VerticalLayout } from '@/workers/layouts/vertical'
+import { RunGraphNode } from "@/models"
+import { ClientLayoutMessage } from '@/workers/runGraph'
+
+type NodeShoveDirection = 1 | -1
+type NearestParentLayoutItem = {
+  row: number,
+  startX: number,
+  endX: number,
+  nextDependencyShove?: NodeShoveDirection
+}
+type NearestParentLayout = Map<string, NearestParentLayoutItem>
+
+export async function getVerticalNearestParentLayout(message: ClientLayoutMessage, horizontal: HorizontalLayout): Promise<VerticalLayout> {
+  const defaultNearestParentPosition = 0
+  const minimumNodeEdgeGap = 16
+
+  const nearestParentLayout: NearestParentLayout = new Map()
+  const layout: VerticalLayout = new Map()
+
+  for (const [nodeId] of message.data.nodes) {
+    const node = message.data.nodes.get(nodeId)
+    const nodeWidth = message.widths.get(nodeId)
+
+    if (!node || !node.start_time || !nodeWidth) {
+      console.warn('NearestParentLayout: Node ID was not found, has no start time or no width', nodeId)
+      continue
+    }
+
+    const startX = horizontal.get(nodeId)!
+    const endX = startX + nodeWidth
+
+    const row = await getNearestParentPosition(node, startX)
+
+    nearestParentLayout.set(nodeId, {
+      row,
+      startX,
+      endX,
+    })
+  }
+
+  for (const [nodeId] of nearestParentLayout) {
+    const { row } = nearestParentLayout.get(nodeId)!
+    layout.set(nodeId, row)
+  }
+
+  purgeNegativeLayoutPositions()
+
+  return layout
+
+  async function getNearestParentPosition(node: RunGraphNode, nodeStartX: number): Promise<number> {
+    // if one dependency
+    if (node.parents && node.parents.length === 1) {
+      if ( nearestParentLayout.has(node.parents[0].id)) {
+        const parent = nearestParentLayout.get(node.parents[0].id)!
+        return await placeNearUpstreamNode(parent, nodeStartX)
+      }
+
+      console.warn('NearestParentLayout: Parent node not found in layout', node.parents[0].id)
+      return defaultNearestParentPosition
+    }
+
+    // if more than one dependency – add to the middle of upstream dependencies
+    if (node.parents && node.parents.length > 0) {
+      const parentLayoutItems = node.parents
+        .map(edge => nearestParentLayout.get(edge.id))
+        .filter((layoutItem: NearestParentLayoutItem | undefined): layoutItem is NearestParentLayoutItem => !!layoutItem)
+
+      const upstreamRows = parentLayoutItems.map(layoutItem => layoutItem.row)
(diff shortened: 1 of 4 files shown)


src/workers/layouts/nearestParentVertical.ts
```diff
@@ -0,0 +1,444 @@
...
+  //   return layoutKeys.length > 0 && layoutKeys.some((nodeId) => {
+  //     const layoutItem = layout[nodeId]
+  //     return isNodesOverlapping({
+  //       firstNodeEndX: layoutItem.endX,
+  //       firstNodePosition: layoutItem.position,
+  //       lastNodeStartX: nodeStartX,
+  //       lastNodePosition: position,
+  //     })
+  //   })
+  // }
```
pleek91, line 165: Do we need to keep this?

src/workers/layouts/nearestParentVertical.ts
```diff
@@ -0,0 +1,444 @@
...
+    lastNodePosition,
+  }: IsNodesOverlappingProps): boolean {
+    return firstNodePosition === lastNodePosition
+      && firstNodeEndX + minimumNodeEdgeGap >= lastNodeStartX
+  }
+
+  function getOverlappingNodeIds(nodeStartX: number, position: number): string[] | undefined {
+    const overlappingNodeIds: string[] = []
+
+    layout.forEach((row, nodeId) => {
```
pleek91, line 246: Can you convert this to a `for...of` loop

src/workers/layouts/nearestParentVertical.ts
```diff
@@ -0,0 +1,444 @@
...
+
+    return nodeStartX + nodeWidth
+  }
+
+  function purgeNegativeLayoutPositions(): void {
+    const allRows = Array.from( layout.values() )
+    const lowestRow = Math.min(...allRows)
+
+    if (lowestRow < 0) {
+      layout.forEach((row) => {
```
pleek91, line 439: `for...of`

src/workers/layouts/nearestParentVertical.ts
```diff
@@ -0,0 +1,444 @@
...
+      console.warn('NearestParentLayout: Node was not found in the horizontal layout and/or widths', nodeId)
+      return 0
+    }
+
+    return nodeStartX + nodeWidth
+  }
+
+  function purgeNegativeLayoutPositions(): void {
+    const allRows = Array.from( layout.values() )
+    const lowestRow = Math.min(...allRows)
```
pleek91, line 436: One micro optimization we could make for this is to track the lowest row as we go like `lowestRow = Math.min(lowestRow, newRow)`. That way we don't have to convert a map of potentially 10k values into an array here.

src/workers/layouts/vertical.ts
```diff
@@ -1,11 +1,13 @@
...
 
-type VerticalLayout = Map<string, number>
+export type VerticalLayout = Map<string, number>
 
-export function getVerticalLayout(message: ClientLayoutMessage, horizontal: HorizontalLayout): VerticalLayout {
+export async function getVerticalLayout(message: ClientLayoutMessage, horizontal: HorizontalLayout): Promise<VerticalLayout> {
   if (message.verticalSettings.mode === 'nearest-parent') {
-    return getVerticalNearestParentLayout(message, horizontal)
+    const layout = await getVerticalNearestParentLayout(message, horizontal)
+    return layout
```
pleek91, line 10:
```suggestion
    return await getVerticalNearestParentLayout(message, horizontal)
```

src/workers/layouts/nearestParentVertical.ts
```diff
@@ -0,0 +1,444 @@
...
+
+      // competitor has equal above and below, or node has more below
+      // place node below competitor
+      nodeShoveRecords.set(competingNodeId, 1)
+    }
+
+    return await placeNearUpstreamNode(competingNodeId, nodeStartX)
+  }
+
+  function getNodeParentDirectionCounts(nodeId: string): [number, number] {
```
pleek91, line 393: nit: a named tuple would help with readability here
```suggestion
  function getNodeParentDirectionCounts(nodeId: string): [up: number, down: number] {
```

src/workers/layouts/nearestParentVertical.ts
```diff
@@ -0,0 +1,444 @@
...
+
+      if (competitorAboveConnections > competitorBelowConnections) {
+        // competitor has more above than below
+        // node wins, shove competitor up
+        await shove({
+          direction: -1,
+          nodeStartX,
+          desiredRow,
+        })
+        return desiredRow
```
pleek91, line 382: Seems like every if statement returns a the same shove and desiredRow but with a different direction. Would help readability to abstract the direction into a utility like `const direction = getShoveDirection(...args)`

src/workers/layouts/nearestParentVertical.ts
```diff
@@ -0,0 +1,444 @@
...
+      const itemAEndX = getNodeEndX(itemAId)
+      const itemBEndX = getNodeEndX(itemBId)
+      if (itemAEndX < itemBEndX) {
+        return 1
+      }
+      if (itemAEndX > itemBEndX) {
+        return -1
+      }
+      return 0
+    })
```
pleek91, line 282: Curious why these needs to be sorted

src/workers/layouts/nearestParentVertical.ts
```diff
@@ -0,0 +1,444 @@
...
+
+      if (isPositionTaken(nodeStartX, row)) {
+        const overlappingNodes = getOverlappingNodeIds(
+          nodeStartX,
+          row,
+        )!
+
+        const overlappingParentNodes = overlappingNodes.filter(layoutId => {
+          return node.parents?.some(({id}) => id === layoutId)
+        })
```
pleek91, line 75: This is a potentially expensive operation because we're calling some within a filter. Maybe we could set the data up differently to make this a one shot calculation

pleek91 (approved):
Made a few comments. Feel free to address in the PR or we can tackle in a follow up.

src/workers/layouts/nearestParentVertical.ts
brandonreid: Curious what you think of the solution I applied for this. Basically assigning a node row to the layout uses a function now instead of simply `layout.set()` so that the lowest row check can be made.

PR #264, opened 2023-10-23
Add the flow run toggle button

pleek91:
# Description
Adds the arrow toggle button to flow run nodes. If it can its fit inside the nodes bar. Otherwise its is positioned to the right of the bar. Same as it is today. 

It has a slightly different style than it does today. We can tweak/change this. Specifically adding a border to it when hovering. I just didn't want to make this PR to large. 

Normal
image
Hovering inside - Different than what we have today. Not because I think this is necessarily better but it is much simpler. We can update this later if we want to. 
image
Hovering outside - This is where we really need that border (coming soon)
image

### DIFF
--- a/src/factories/nodeArrowButton.ts
+++ b/src/factories/nodeArrowButton.ts
@@ -1,0 +1,58 @@
+import { Container, ColorMatrixFilter } from 'pixi.js'
+import { ArrowDirection, ArrowStyle, arrowFactory } from '@/factories/arrow'
+import { BarStyle, barFactory } from '@/factories/bar'
+
+type NodeArrowBarStyles = {
+  arrow: Omit<ArrowStyle, 'rotate'>,
+  button: BarStyle,
+  isOpen: boolean,
+}
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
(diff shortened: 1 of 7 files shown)


src/factories/nodeArrowButton.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    return container
+  }
+
+  function onMouseover(): void {
+    filter.brightness(0.5, false)
+  }
+
+  function onMouseout(): void {
+    filter.brightness(1, false)
```
brandonreid, line 51: I wonder if filter.brightness() is more efficient than arrow.alpha?

brandonreid (approved):

src/factories/nodeArrowButton.ts
pleek91: Not sure what you mean

brandonreid: So any sprite/graphics/etc. has an `alpha` value. Rather than a filter here, you could do `container.alpha = 0.5`

brandonreid: ah gotchya, makes sense!

PR #266, opened 2023-10-24
Implement culling of nodes

pleek91:
# Description
Not rendering nodes outside of the viewport means rendering should be faster when zoomed in.


src/objects/viewport.ts
```diff
@@ -62,8 +63,10 @@ export async function centerViewport({ animate }: CenterViewportParameters = {})
     throw new Error('Nodes container not found')
   }
 
-  // when we get to culling we might need to turn it off for this measurement
+  pauseCulling()
   const { x, y, width, height } = container.getLocalBounds()
+  resumeCulling()
+
```
pleek91, line 69: I think we could probably do this much more efficiently. But we're using this in v1 pretty successfully so not gong to optimize this prematurely.

brandonreid (approved):

PR #267, opened 2023-10-24
Cull labels when not readable

pleek91:
# Description
Makes labels not visible when the viewport is lower than a 0.2 scale. At this scale labels are not readable. Helps increase performance when zoomed out and many labels would be rendered.


src/factories/nodeFlowRun.ts
```diff
@@ -59,8 +59,8 @@ export async function flowRunContainerFactory(node: RunGraphNode) {
     ])
 
     nodesContainer.visible = true
+    nodesContainer.once('rendered', () => cull())
```
pleek91, line 62: Noticed this wasn't working before. Need to wait until AFTER everything is rendered before calling `cull`

This made me realize this is highly inefficient to render things and then cull them. But its working for now. It would be much more efficient to not render things initially and then render them if they are visible. Think we could accomplish that by setting `renderable = false` on nodes when creating them (before adding them as children). Might mess with that later but for now this works well. Just a bad fps drop when you open a sub node with a lot of nodes because they must all be rendered and then once they are culled the fps jumps back up.

brandonreid: Note that when a flow is first drawn, all nodes are made visible. I guess it does matter for labels in this case though.

brandonreid (approved):

src/factories/nodeFlowRun.ts
pleek91: Right, its rendered and then culled. Which is my point here. Unless you're getting at something else?

brandonreid: Na at first I thought you may be thinking about the nodes, which will all be visible at first anyway. Ignore me, haha.

PR #268, opened 2023-10-25
Node edges

pleek91:
# Description
Addes edges to nodes

(note this screenshot shows a bug in the nearest parent layout that we're aware of)
image

image


src/factories/cap.ts
pleek91, line 1: Unrelated to edges but I did a little organization of textures. All textures and their subscription wrappers are in `/textures`

brandonreid (approved):

PR #269, opened 2023-10-26
use on-the-fly bitmap font spritesheets

brandonreid:
This uses the v1 approach for loading bitmap fonts and creating a sprite sheet from them. This approach is ideal for text at scale. It has a fixed fidelity, so it will be blurry when zoomed in really far, but the rendered image textures for each characters are far more performant than vector text which would normally be suitable for small amounts of text in 2D games.

Note: I haven't played with using `tint` for coloring text, but if it doesn't work, we will need to track the various text styles we need to use. I've named the first one `nodeTextDefault` to leave room for `nodeTextInverse` (which will be needed for text outside of a node in light mode).

Also centers the labels vertically.


src/factories/nodeFlowRun.ts
```diff
@@ -130,10 +130,10 @@ export async function flowRunContainerFactory(node: RunGraphNode) {
 
     const labelMinLeft = Math.max(barRight, buttonRight)
     const inside = barWithoutMargin > labelMinLeft + label.width
-    const y = bar.height / 2
+    const yOffset = 0.5
+    const y = bar.height / 2 - label.height / 2 - yOffset
     const x = inside ? labelMinLeft + margin : arrowButton.x + arrowButton.width + margin
 
-    label.anchor.set(0, 0.5)
```
brandonreid, line 136: I couldn't quite understand where this spacing was coming from, but it performed better as an offset on the y for some reason.

pleek91 (approved):

src/factories/nodeFlowRun.ts
pleek91: yeah I was trying to center by setting the anchor to 0.5 and then positioning to bar.height / 2. But with a predictable line height this probably doesn't work quite right. (Wasn't quite right to begin with).

PR #275, opened 2023-10-26
speed up nearestParent layout

brandonreid:
This PR uses a `rowTracker` map to track which nodes are in which row. This significantly improves checks for overlaps.

Here's the time it took to calculate the nearestParent layout on 2k nodes before and after this PR:
Before: **28,686.7 ms**
After: **1,394.1 ms**


pleek91 (approved):
Yeah boi!

PR #276, opened 2023-10-26
Add borders to sub nodes

pleek91:
# Description

Before
image
image

After
image
image


src/factories/border.ts
pleek91, line 1: Changes to this file fix an issue where if the width/height of the border was smaller than the radius the sprites would overlap. This makes sure the radius is not larger than the width/height.

brandonreid (approved):
Nit here but g2g

PR #277, opened 2023-10-27
add animation to nodes and edges

brandonreid:
Animates updates to nodes and edge positions.

### DIFF
--- a/src/factories/animation.ts
+++ b/src/factories/animation.ts
@@ -1,13 +1,18 @@
 import gsap from 'gsap'
 import { waitForConfig } from '@/objects/config'
+import { pauseCulling, resumeCulling } from '@/objects/culling'
 
 export async function animate(el: gsap.TweenTarget, vars: gsap.TweenVars, skipAnimation?: boolean): Promise<gsap.core.Tween> {
   const config = await waitForConfig()
   const duration = skipAnimation ? 0 : config.animationDuration / 1000
 
+  pauseCulling()
+
   return gsap.to(el, {
     ...vars,
     duration,
     ease: 'power1.out',
+  }).then(() => {
+    resumeCulling()
   })
 }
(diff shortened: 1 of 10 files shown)


src/objects/culling.ts
```diff
@@ -55,6 +55,11 @@ export async function cull(): Promise<void> {
 }
 
 export async function resumeCulling(): Promise<void> {
+  const application = await waitForApplication()
+
+  if (cullInstance) {
+    cullInstance.cull(application.renderer.screen)
+  }
```
pleek91: This function needs to be removed. There isn't actually a concept of "pause and resume" There's just culled and unculled. So we should rename the `pauseCulling` function `uncull` and then call `uncull` and `cull`.

src/factories/animation.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  pauseCulling()
+
+  return gsap.to(el, {
+    ...vars,
+    duration,
+    ease: 'power1.out',
+  }).then(() => {
+    resumeCulling()
+  })
```
pleek91: Couldn't we just await to `gsap.to` and then call cull? 

And just make this return `Promise<void>`. I don't think returning the tween is actually useful.

src/factories/edge.ts
```diff
@@ -10,6 +11,9 @@ import { repeat } from '@/utilities/repeat'
 
 export type EdgeFactory = Awaited<ReturnType<typeof edgeFactory>>
 
+const arrowOffset = 8
+const edgeTargetOffset = 2
+
```
pleek91, line 16: Thank you for being a better dev and removing my ugly comments 😂 ❤️

pleek91 (approved):

src/factories/animation.ts
pleek91: not relevant anymore

src/objects/culling.ts
brandonreid: Went ahead and renamed pauseCulling to uncull and updated. Removed resumeCull. 👍

PR #279, opened 2023-10-27
Row and column gaps

pleek91:
# Description
Adds some configurable gap between rows and columns. This is most noticeable when looking at the dependency view. Also moves some row/column logic to the worker so the layout response is more useful. 

Before
image

After
image

### DIFF
--- a/src/factories/nodes.ts
+++ b/src/factories/nodes.ts
@@ -1,5 +1,5 @@
 import { Container } from 'pixi.js'
-import { DEFAULT_NODES_CONTAINER_NAME, DEFAULT_POLL_INTERVAL } from '@/consts'
+import { DEFAULT_LINEAR_COLUMN_SIZE_PIXELS, DEFAULT_NODES_CONTAINER_NAME, DEFAULT_POLL_INTERVAL } from '@/consts'
 import { EdgeFactory, edgeFactory } from '@/factories/edge'
 import { NodeContainerFactory, nodeContainerFactory } from '@/factories/node'
 import { offsetsFactory } from '@/factories/offsets'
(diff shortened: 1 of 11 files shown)


src/factories/nodes.ts
```diff
@@ -160,12 +168,16 @@ export async function nodesContainerFactory(runId: string) {
...
+    }
+
     for (const [nodeId, node] of nodes) {
-      const position = layout.get(nodeId)
+      const position = nodesLayout.positions.get(nodeId)
 
       if (!position) {
         console.warn(`Could not find node in layout: Skipping ${nodeId}`)
-        return
+        continue
```
pleek91, line 180: I found a few places we were returning in a for loop rather than continuing. Unrelated to the feature work but didn't make a separate PR.

src/factories/nodes.ts
```diff
@@ -213,33 +229,44 @@ export async function nodesContainerFactory(runId: string) {
...
+    return rows.getTotalValue(nodesLayout.maxRow)
+  }
+
+  function getWidth(): number {
+    if (!nodesLayout) {
+      return 0
+    }
 
-    return height
+    throw new Error('Not implemented')
```
pleek91, line 269: Placeholder. I move maxRow into the layout worker and realized that maxColumn would be helpful for calculating the width of a run. But I'll finish that in a future PR.

src/factories/offsets.ts
pleek91, line 1: Made this factory agnostic of the y axis and things like node height. Was going to use it for the columns as well but didn't work out like I'd hope. But felt this was still a good change to keep this modular.

znicholasbrown (approved):

PR #282, opened 2023-10-30
Fix incorrect layout when objects are culled

pleek91:
# Description
Because various parts of the layout depend on the physical size of objects its important those objects bounds are calculated correctly. When pixi calculates the bounds of a container it skips any children that have their `visible` or `renderable` properties set to false. Which means if a node or label is culled and a layout calculation is requested, the layout will be calculated using the wrong dimensions. 

To fix this I'm extending pixi's built in `Container` class with a custom `BoundsContainer` and using this new custom class for node containers. This custom implementation is exactly the same as the pixi version EXCEPT it does not skip any children even if their `visible` or `renderable` properties are set to false. This means that the bounds is always accurate. 

Pixi still skips these children when rendering. So we still get the majority (if not all) of the performance boost of culling. From my testing the performance impact of this change was unnoticeable with 2k nodes. And it fixed the layout problem

## Reproduction
1. Use the dependency view
2. Zoom out until labels dissapear
3. Switch to trace view
4. Recenter the viewport

Before
image

After
image

### DIFF
--- a/src/factories/nodeFlowRun.ts
+++ b/src/factories/nodeFlowRun.ts
@@ -1,10 +1,10 @@
-import { Container } from 'pixi.js'
 import { DEFAULT_NODE_CONTAINER_NAME } from '@/consts'
 import { borderFactory } from '@/factories/border'
 import { nodeLabelFactory } from '@/factories/label'
 import { nodeArrowButtonFactory } from '@/factories/nodeArrowButton'
 import { nodeBarFactory } from '@/factories/nodeBar'
 import { nodesContainerFactory } from '@/factories/nodes'
+import { BoundsContainer } from '@/models/boundsContainer'
 import { NodeSize } from '@/models/layout'
 import { RunGraphNode } from '@/models/RunGraph'
 import { waitForConfig } from '@/objects/config'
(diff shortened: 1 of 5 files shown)


src/factories/nodeFlowRun.ts
```diff
@@ -81,13 +80,13 @@ export async function flowRunContainerFactory(node: RunGraphNode) {
 
   async function open(): Promise<void> {
     isOpen = true
+    container.addChild(nodesContainer)
```
pleek91, line 83: Worth calling out. Since the node is a BoundsContainer if the container for the sub nodes is added it will be part of the bounds even if it is `visible = false` which is what we were doing before. 

So adding and removing the nodes container when opening and closing. I think this is fine because its only happening when a click is triggered when we're rendering the node and its children anyway. If this was happening in a ticker I think this would be a problem. But I think this is fine.

pleek91: We could add a custom flag that BoundsContainer would check to ACTUALLY skip. But that seemed messier than this and this seemed to work well.

src/factories/nodes.ts
```diff
@@ -45,7 +46,12 @@ export async function nodesContainerFactory(runId: string) {
 
   container.name = DEFAULT_NODES_CONTAINER_NAME
 
-  emitter.on('layoutUpdated', () => renderNodes())
+  emitter.on('layoutUpdated', () => {
+    nodesLayout = null
+    rows.clear()
+    columns.clear()
+    renderNodes()
+  })
```
pleek91, line 54: This also fixed some issues where row offsets from one view would carry over when the view was switched. Seems like a good idea anyway to clear things out when the view is changed.

src/models/boundsContainer.ts
pleek91, line 1: Found this technique suggested in a pixi issue. I've tried half a dozen other things and this was the only one that actually worked. 

https://github.com/pixijs/pixijs/issues/4338#issuecomment-333294015

znicholasbrown (approved):

PR #289, opened 2023-10-31
Refactor how run data is fetched and rendered

pleek91:
# Description
Previously the nodesContainerFactory was responsible for fetching and rendering its data. This was different than the rest of the factories but worked well up until now. But in order to set the initial horizontal scale we need to know how long the run was. So pulling out the fetching of data from the rendering of data.

### DIFF
--- a/src/factories/data.ts
+++ b/src/factories/data.ts
@@ -1,0 +1,33 @@
+import { DEFAULT_POLL_INTERVAL } from '@/consts'
+import { RunGraphData } from '@/models/RunGraph'
+import { waitForConfig } from '@/objects/config'
+
+type DataCallback = (data: RunGraphData) => void
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export async function dataFactory(runId: string, callback: DataCallback) {
+  const config = await waitForConfig()
+
+  let interval: ReturnType<typeof setInterval> | undefined = undefined
+
+  async function start(): Promise<void> {
+    const data = await config.fetch(runId)
+
+    callback(data)
+
+    if (!data.end_time) {
+      interval = setTimeout(() => start(), DEFAULT_POLL_INTERVAL)
+    }
+  }
+
+  // todo: need a global way of stopping this when the graph is stopped
+  function stop(): void {
+    clearInterval(interval)
+  }
+
+  return {
+    start,
+    stop,
+  }
+
+}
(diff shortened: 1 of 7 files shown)


src/factories/data.ts
pleek91, line 1: New factory for fetching data. Provide a runId and a data callback and it returns a start and stop method.

src/factories/nodeFlowRun.ts
pleek91, line 1: Since the nodesContainerFactory doesn't fetch its own data now. The flow run needs to fetch data and provide it.

src/factories/nodes.ts
pleek91, line 1: now this factory expects data to be provided when calling the `render` method same as all the other factories. It does store a reference to the provided data so that if the layout settings change it can automatically re-render. The rest of the functionality is mostly the same. Little reworking and renaming.

src/objects/events.ts
```diff
@@ -25,6 +25,7 @@ type Events = {
   fontsLoaded: Fonts,
   containerCreated: Container,
   layoutUpdated: LayoutSettings,
+  layoutCreated: LayoutSettings,
```
pleek91, line 28: Previously the layout had defaults for everything so it was never "created". But now the layout settings emit a created event when the horizontal scale is initialized.

src/objects/nodes.ts
pleek91, line 1: Now this object fetches run data and passes it to the nodes factory to render data.

src/objects/settings.ts
```diff
@@ -2,12 +2,23 @@ import { addSeconds } from 'date-fns'
...
 import { DEFAULT_LINEAR_COLUMN_SIZE_PIXELS, DEFAULT_TIME_COLUMN_SIZE_PIXELS, DEFAULT_TIME_COLUMN_SPAN_SECONDS } from '@/consts'
 import { HorizontalMode, LayoutSettings, VerticalMode } from '@/models/layout'
-import { emitter } from '@/objects/events'
+import { EventKey, emitter, waitForEvent } from '@/objects/events'
+import { waitForRunData } from '@/objects/nodes'
+
+export async function startSettings(): Promise<void> {
+  const data = await waitForRunData()
+
+  setHorizontalScale(1)
```
pleek91, line 11: Now we have a place where we can set the initial horizontal scale. Next PR will set this based on the length of the run data

brandonreid (approved):

PR #294, opened 2023-10-31
Add node selection

pleek91:
# Description
Adds the ability to click on any node and select it. This is hooked up to a v-model on the component so it can be controlled externally. 

image

### DIFF
--- a/src/components/RunGraph.vue
+++ b/src/components/RunGraph.vue
@@ -10,22 +10,25 @@
 </template>
 
 <script lang="ts" setup>
-  import { computed, onBeforeUnmount, onMounted, ref } from 'vue'
+  import { computed, onBeforeUnmount, onMounted, ref, watch } from 'vue'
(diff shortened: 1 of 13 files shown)


src/components/RunGraph.vue
```diff
@@ -41,6 +44,10 @@
     },
   })
 
+  watch(() => props.selected, selected => selectNode(selected))
```
znicholasbrown, line 47: you're ok with this firing twice for each selection because of the escape hatch in `selectNode`, right?

znicholasbrown (approved):

src/components/RunGraph.vue
pleek91: Correct, generally I'd have that check here but wanted that to happen no matter what triggered it.

znicholasbrown: makes sense

PR #295, opened 2023-11-01
Add settings and thresholds for guides and animations

pleek91:
# Description
Adds config options for setting thresholds for disabling animations and guides. This helps with performance on larger flows. Currently animations are disabled on flows with 500 nodes or more (same as v1) and edges are disabled on flows with 100 nodes or more (vs 40 in v1).

Edges can always be toggled manually by the user in the graph settings
image


src/factories/nodes.ts
```diff
@@ -25,6 +25,7 @@ export async function nodesContainerFactory() {
   const nodes = new Map<string, NodeContainerFactory>()
   const edges = new Map<EdgeKey, EdgeFactory>()
   const container = new Container()
+  const edgesContainer = new Container()
```
pleek91, line 28: Adding edges to a dedicated container so they can be added/removed easily

src/factories/nodes.ts
```diff
@@ -82,6 +83,15 @@ export async function nodesContainerFactory() {
...
 
   async function createEdges(data: RunGraphData): Promise<void> {
+    const settings = await waitForSettings()
+
+    if (settings.disableEdges) {
+      container.removeChild(edgesContainer)
+      return
+    }
+
+    container.addChildAt(edgesContainer, 0)
```
pleek91, line 93: `adChildAt` means edges will always be behind the rest of the node content.

brandonreid: Heads up, in V1, a transparent rectangle is added to the viewport and also set behind nodes as a deselect target when a node is selected. Might be worth tracking these top level z-indexes (can do later of course)

brandonreid (approved):

src/factories/nodes.ts
pleek91: I'm hoping not to use sortable children since sorting large arrays is very slow. As long as we add them in the correct order we shouldn't need to use z-indexes. 

Also I'm using the viewport itself as the deselect target and using stopPropogation on the nodes themselves.

PR #298, opened 2023-11-01
Bug fixes from implementation

pleek91:
# Description
Various bug fixes. See comments for details

### DIFF
--- a/demo/sections/components/RunGraphDemo.vue
+++ b/demo/sections/components/RunGraphDemo.vue
@@ -37,15 +37,15 @@
 
   // just hard coding the values here for now. these will come from ui-library
   const stateTypeColors = {
-    completed: '#219D4B',
-    running: '#09439B',
-    scheduled: '#E08504',
-    pending: '#554B58',
-    failed: '#DE0529',
-    cancelled: '#333333',
-    cancelling: '#333333',
-    crashed: '#EA580C',
-    paused: '#554B58',
+    COMPLETED: '#219D4B',
+    RUNNING: '#09439B',
+    SCHEDULED: '#E08504',
+    PENDING: '#554B58',
+    FAILED: '#DE0529',
+    CANCELLED: '#333333',
+    CANCELLING: '#333333',
+    CRASHED: '#EA580C',
+    PAUSED: '#554B58',
   } satisfies Record<StateType, string>
 
   const config: RunGraphConfig = {
--- a/src/factories/data.ts
+++ b/src/factories/data.ts
@@ -9,13 +9,19 @@ export async function dataFactory(runId: string, callback: DataCallback) {
   const config = await waitForConfig()
 
   let interval: ReturnType<typeof setInterval> | undefined = undefined
+  let data: RunGraphData | null = null
 
   async function start(): Promise<void> {
-    const data = await config.fetch(runId)
+    try {
+      data = await config.fetch(runId)
 
-    callback(data)
+      callback(data)
+    } catch (error) {
+      console.error(error)
+    }
 
-    if (!data.end_time) {
+
+    if (data && !data.end_time) {
       interval = setTimeout(() => start(), DEFAULT_POLL_INTERVAL)
     }
   }
--- a/src/factories/edge.ts
+++ b/src/factories/edge.ts
@@ -26,6 +26,7 @@ export async function edgeFactory() {
   const pixel = await getPixelTexture()
   const points = repeat(DEFAULT_EDGE_POINTS, () => new Point())
   const rope = new SimpleRope(pixel, points)
+
   let initialized = false
 
   container.name = DEFAULT_EDGE_CONTAINER_NAME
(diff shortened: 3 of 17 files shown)


demo/sections/components/RunGraphDemo.vue
pleek91, line 1: When implementing the actual api in staging realized the endpoint was returning uppercase types. And not mapping these. So updated the types this repo is expecting.

src/factories/data.ts
pleek91, line 1: Fixes the data factory from breaking if there is an intermittent error.

src/factories/edge.ts
pleek91, line 1: Skipping animation based on if the edge is initialized. Previously was skipping animations only if the run was initialized. Which meant new edges animated in.

src/factories/node.ts
```diff
@@ -14,12 +14,15 @@ export type NodeContainerFactory = Awaited<ReturnType<typeof nodeContainerFactor
 
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
 export async function nodeContainerFactory(node: RunGraphNode) {
+  const application = await waitForApplication()
   const cull = await waitForCull()
   const { animate } = await animationFactory()
   const { element: container, render: renderNode, bar } = await getNodeFactory(node)
-  const cacheKey: string | null = null
 
+  let internalNode = node
```
pleek91, line 22: Keeping an internal value of the node that gets updated every time render is called. This is used to re-render internally.

src/factories/node.ts
```diff
@@ -14,12 +14,15 @@ export type NodeContainerFactory = Awaited<ReturnType<typeof nodeContainerFactor
...
+  let cacheKey: string | null = null
```
pleek91, line 23: Cache key was never getting set before so every render call was a cache miss. Actually setting the cacheKey now and fixed the logic for creating a cache key. Previously was caching on `end_date.toString()` which only had second precisions. The cache key is now just based on the `state_type` and `end_date.getTime()` values.

src/factories/node.ts
```diff
@@ -14,12 +14,15 @@ export type NodeContainerFactory = Awaited<ReturnType<typeof nodeContainerFactor
...
   let nodeIsSelected = false
+  let initialized = false
```
pleek91, line 25: Same as for edges, used to skip animations on first render.

src/factories/node.ts
```diff
@@ -29,34 +32,54 @@ export async function nodeContainerFactory(node: RunGraphNode) {
...
+  }
+
+  function stopTicking(): void {
+    application.ticker.remove(tick)
+  }
+
+  function tick(): void {
+    render(node)
+  }
+
```
pleek91, line 82: Previous ticker implementation would just keep ticking because it would call render with the same value. So even after the node entered completed it would keep ticking. Nothing visible happened but waste of processing.

src/factories/node.ts
```diff
@@ -72,21 +95,18 @@ export async function nodeContainerFactory(node: RunGraphNode) {
...
-    const keys = Object.keys(node).sort((keyA, keyB) => keyA.localeCompare(keyB)) as (keyof RunGraphNode)[]
-    const values = keys.map(key => {
-      const value = node[key] ?? new Date()
-
-      return value.toString()
-    })
+    const endTime = node.end_time ?? new Date()
 
-    return values.join(',')
+    return `${node.state_type},${endTime.getTime()}`
```
pleek91, line 100: Must simpler cache key.

src/factories/nodeBar.ts
pleek91, line 1: Just fixing this logic to reflect the uppercase state types

src/factories/nodeFlowRun.ts
```diff
@@ -25,6 +25,7 @@ export async function flowRunContainerFactory(node: RunGraphNode) {
     renderNodes(data)
   })
 
+  let internalNode = node
```
pleek91, line 28: again keeping an internal reference for re-rendering.

src/factories/nodeFlowRun.ts
```diff
@@ -44,11 +45,13 @@ export async function flowRunContainerFactory(node: RunGraphNode) {
...
+  async function render(node: RunGraphNode): Promise<BoundsContainer> {
+    internalNode = node
+
     await renderBar(node)
-    await renderArrowButton()
-    await renderLabel()
-    await renderBorder()
+    await renderArrowButton(node)
+    await renderLabel(node)
+    await renderBorder(node)
```
pleek91, line 54: realized these were all rendering based on the node passed into the factory and not the value passed into the render function.

src/factories/nodes.ts
pleek91, line 1: Don't care about initialized or animation here now because edges and nodes track their own initialized values.

src/models/index.ts
pleek91, line 1: Needed to export some types

src/models/selection.ts
pleek91, line 1: Realized selection needed to emit the kind and not just the id.

src/objects/culling.ts
```diff
@@ -17,25 +15,18 @@ export async function startCulling(): Promise<void> {
     toggle: 'renderable',
   })
 
-  callback = (): void => {
+  application.ticker.add(() => {
```
pleek91, line 18: Replaced all `Ticker.shared.add` calls with `application.ticker.add` because the shared ticker keeps ticking even after the application is torn down. The application ticker automatically stops when the application is destroyed.

src/workers/runGraph.ts
pleek91, line 1: worker was erroring trying to load in nebula-ui. inlining it keeps it internal to this package.

brandonreid (approved):

PR #300, opened 2023-11-02
Add guide lines

brandonreid:


src/consts.ts
```diff
@@ -11,4 +11,6 @@ export const DEFAULT_TEXTURE_RESOLUTION = 10
 export const DEFAULT_EDGE_POINTS = 20
 export const DEFAULT_EDGE_MINIMUM_BEZIER = 64
 export const DEFAULT_HORIZONTAL_SCALE_MULTIPLIER = 0.25
-export const DEFAULT_HORIZONTAL_SCALE = 1
\ No newline at end of file
+export const DEFAULT_HORIZONTAL_SCALE = 1
+export const DEFAULT_GUIDES_COUNT = 30
```
brandonreid, line 15: I'm thinking I may not need 30.. unless the user has a large screen. Might be worth calculating an amount instead of using a fixed one.

brandonreid: Could also just create them on the fly when I need them so there'll only ever be as many as neccesary..

src/factories/guides.ts
```diff
@@ -0,0 +1,160 @@
...
+      pauseGuides()
+      return
+    }
+
+    resumeGuides()
+  }
+
+  function pauseGuides(): void {
+    paused = true
+    element.visible = false
```
brandonreid, line 147: Thinking `visible` is the way to go over `renderable` just based on these descriptions:

`visible`
image

`renderable`
image

src/factories/guide.ts
```diff
@@ -0,0 +1,70 @@
...
+import { waitForScale } from '@/objects/scale'
+
+export type GuideFactory = Awaited<ReturnType<typeof guideFactory>>
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export async function guideFactory() {
+  const application = await waitForApplication()
+  const viewport = await waitForViewport()
+  const cull = await waitForCull()
+  const scale = await waitForScale()
```
pleek91, line 18: Since the scale can change when the layout changes you'll wanna get the scale when you use it rather than here in the factory.

If you change the horizontal scale you'll see the guides don't update

src/factories/guide.ts
```diff
@@ -0,0 +1,70 @@
...
+  const { styles } = await waitForConfig()
```
pleek91, line 19: Don't destructure the config. Otherwise if it changes you won't get the latest value when accessing styles

src/factories/guide.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  async function renderLabel(date: Date): Promise<void> {
+    label.text = currentLabelFormatter(date)
+    label.fontSize = styles.guideTextSize
+    label.tint = styles.guideTextColor
+    label.position.set(styles.guideTextLeftPadding, styles.guideTextTopPadding)
+
+    await label
+  }
```
pleek91, line 64: seems like this doesn't need to be async

src/factories/guides.ts
```diff
@@ -0,0 +1,160 @@
...
+import { waitForScale } from '@/objects/scale'
+import { repeat } from '@/utilities/repeat'
+import { formatDateFns, labelFormats, timeIncrements } from '@/utilities/timeIncrements'
+
+const visibleGuideBoundsMargin = 300
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export async function guidesFactory() {
+  const viewport = await waitForViewport()
+  const scale = await waitForScale()
```
pleek91, line 17: Same comment about the scale not getting updated if you get it here

src/factories/guides.ts
pleek91: currently we're using `renderable` for the global culler and `visible` for everything else 👍

src/factories/guides.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  function getGuideTimes(): number[] {
+    return repeat(DEFAULT_GUIDES_COUNT, (index) => {
+      return currentAnchor + currentIncrement * index
+    })
+  }
+
+  function onLayoutUpdate(layout: LayoutSettings): void {
+    if (layout.horizontal !== 'trace') {
```
pleek91, line 137:
```suggestion
    if (layout.isTrace()) {
```

src/factories/guides.ts
```diff
@@ -0,0 +1,160 @@
...
+    guides.set(index, response)
+  }
+
+  function update(): void {
+    if (paused) {
+      return
+    }
+
+    const left = scale.invert(viewport.left - visibleGuideBoundsMargin)
+    const gapDate = scale.invert(viewport.left - visibleGuideBoundsMargin + DEFAULT_GUIDES_MIN_GAP / viewport.scale.x) as Date
```
pleek91, line 64: nit: You're already type guarding `left` might as well type guard `gapDate` as well rather than using an `as`

src/factories/guide.ts
```diff
@@ -0,0 +1,73 @@
...
+import { waitForScale } from '@/objects/scale'
+
+export type GuideFactory = Awaited<ReturnType<typeof guideFactory>>
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export async function guideFactory() {
+  const application = await waitForApplication()
+  const viewport = await waitForViewport()
+  const cull = await waitForCull()
+  const config = await waitForConfig()
```
brandonreid, line 18: @pleek91 in this scenario, can I expect config to always provide the latest values? Or should this be moved into the render functions?

pleek91: since config is an object every time you access it like config.xyz it will always be the correct value. You just cannot destructure it and use the value later or you'll likely end up with the same value each time. Similar to breaking reactivity in vue

brandonreid: Gotchya, appreciate it.

pleek91 (approved):

src/consts.ts
brandonreid: Might experiment with this idea separately but will leave it for now.

PR #302, opened 2023-11-02
add keyboard shortcuts

brandonreid:


src/components/RunGraph.vue
```diff
@@ -70,11 +75,36 @@
       stage: stage.value,
       props,
     })
+
+    window.addEventListener('keydown', keyboardShortcutListener)
```
pleek91, line 79: Can we use `useKeyDown` for this

pleek91 (approved):

PR #304, opened 2023-11-02
Guides tweaks and fix guides missing on initial load

pleek91:
# Description
Fixes
- Guides missing on initial load
- Guide warnings on initial load
- Guides lagging behind nodes (rubber band effect)
- Viewport date range v-model not syncing with the pixi viewport


src/components/RunGraph.vue
pleek91, line 1: Makes sure the pixi viewport gets updated to reflect the date range if its updated externally

src/factories/guide.ts
```diff
@@ -27,23 +30,30 @@ export async function guideFactory() {
   const label = inter('')
   element.addChild(label)
 
+  let scale = await waitForScale()
   let currentDate: Date | undefined
   let currentLabelFormatter: FormatDate
 
+  emitter.on('scaleUpdated', updated => scale = updated)
```
pleek91, line 37: This keeps scales up to date without having to make the functions that depend on them async.

src/factories/guide.ts
```diff
@@ -27,23 +30,30 @@ export async function guideFactory() {
...
+
   application.ticker.add(() => {
+    if (settings.disableGuides) {
+      return
+    }
```
pleek91, line 42: Tracking if guides should be rendered or not externally same as the other things that can be disabled (animations, edges)

src/factories/guide.ts
```diff
@@ -59,8 +69,7 @@ export async function guideFactory() {
     label.position.set(config.styles.guideTextLeftPadding, config.styles.guideTextTopPadding)
   }
 
-  async function updatePosition(): Promise<void> {
-    const scale = await waitForScale()
+  function updatePosition(): void {
```
pleek91, line 72: Making this sync keeps it from lagging behind the viewport.

src/factories/guides.ts
pleek91, line 1: This is reorganized and simplified so you can just call render to update the guides. The guides object is now responsible for updating the guides based on events.

brandonreid (approved):

src/objects/guides.ts
pleek91, line 1: This object is now in charge of any re-renders that need to happen based on the viewport or layout changing.

src/objects/viewport.ts
pleek91, line 1: These changes are related to fixing the viewport v-model

PR #310, opened 2023-11-03
highlight selection paths

brandonreid:
Performing well on 2k nodes:

https://github.com/PrefectHQ/graphs/assets/6776415/9d3946ba-95d8-46c6-8f67-651fe2a5b8ef

### DIFF
--- a/src/factories/nodes.ts
+++ b/src/factories/nodes.ts
@@ -47,19 +47,17 @@ export async function nodesContainerFactory() {
   container.name = DEFAULT_NODES_CONTAINER_NAME
 
   emitter.on('layoutSettingsUpdated', async () => {
-    const selected = getSelected()
-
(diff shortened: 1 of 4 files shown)


src/factories/nodes.ts
```diff
@@ -53,6 +53,14 @@ export async function nodesContainerFactory() {
     }
   })
 
+  emitter.on('nodeSelected', (selection) => {
+    unHighlightPath()
+
+    if (selection?.id && nodes.has(selection.id)) {
+      highlightPath(selection.id)
+    }
+  })
```
pleek91, line 62: This is looping every node and every edge twice but we could do it with a single loop. Rather than unHighlighting lets loop and set the opacity based on if the edge should be highlighted or not. If nothing should be highlighted the pathIds can just be an empty array

pleek91 (approved):

PR #313, opened 2023-11-03
Add playhead

brandonreid:
Playhead ready! There are still issues with running that we'll want to investigate once pending tasks are removed from view:

https://github.com/PrefectHQ/graphs/assets/6776415/74c8136e-a7f9-43d6-8d6e-bfb301d5894e

After refresh:
image

### DIFF
--- a/src/factories/playhead.ts
+++ b/src/factories/playhead.ts
@@ -1,4 +1,3 @@
-import { Container } from 'pixi.js'
 import { rectangleFactory } from '@/factories/rectangle'
 import { waitForApplication, waitForViewport } from '@/objects'
 import { waitForConfig } from '@/objects/config'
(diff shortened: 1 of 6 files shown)


src/factories/playhead.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    if (
+      !viewport.moving
+      && playheadStartedVisible
+      && element.position.x > application.screen.width - autoViewportUpdatePadding
+    ) {
+      const originalLeft = scale.invert(viewport.left)
+
+      viewport.zoomPercent(-0.1, true)
+      viewport.left = scale(originalLeft)
```
pleek91, line 53: aren't these to lines just setting `viewport.left = viewport.left`? 

```
const originalLeft = scale.invert(viewport.left)
viewport.left = scale(originalLeft)
```

src/factories/playhead.ts
```diff
@@ -0,0 +1,66 @@
...
+  const ticker = (): void => {
+    render()
+    checkViewport()
+  }
+
+  function render(): void {
+    playhead.width = config.styles.playheadWidth
+    playhead.height = application.stage.height
+    playhead.tint = config.styles.playheadColor
+    element.position.x = scale(new Date()) * viewport.scale._x + viewport.worldTransform.tx
```
pleek91, line 34: why `viewport.scale._x` and not `viewport.scale.x`?

src/factories/playhead.ts
pleek91: Yeah seems like its the same as this
```
const originalLeft = viewport.left
viewport.zoomPercent(-0.1, true)
viewport.left = originalLeft
```

pleek91:
@brandonreid pushed a commit to simplify this a bit. Double check that this still functions as it should. From my testing it works the same.

src/factories/playhead.ts
brandonreid: Not sure tbh, grabbed from v1 but `.x` seems to work just fine so updated it.

brandonreid:
> @brandonreid pushed a commit to simplify this a bit. Double check that this still functions as it should. From my testing it works the same.

LGTM! Tried it out, functions well. Nice cleanup!

pleek91 (approved):

PR #315, opened 2023-11-05
Bug Fixes

pleek91:
# Description
This PR fixes a few bugs. See the comments for explanations.


src/factories/node.ts
pleek91, line 1: The main change here is adding `layout.horizontalScaleMultiplier` to the cache key so that the nodes re-render when that setting changes.

src/objects/cache.ts
pleek91, line 1: One bug discovered from QA was that after navigating away from a flow run and then viewing the flow run again (or any other graph) everything but the labels would be missing. This was because we're using subscriptions to cache generated textures. But those textures are only valid per application. So when navigating away from a graph and the `stop` method is called, the application gets torn down and destroys the textures. 

So this utility allows using subscriptions that will get torn down automatically when the `stop` method is called. Which fixes the textures issue.

src/objects/index.ts
```diff
@@ -39,23 +40,29 @@ export function start({ stage, props }: StartParameters): void {
...
+    stopFonts()
+    stopCulling()
+    stopLabelCulling()
+    stopEdgeCulling()
+    stopSettings()
+    stopSelection()
+    stopCache()
+  } catch (error) {
+    console.error(error)
+  }
```
pleek91, line 67: Added this while debugging the viewport teardown issue. Even if something errors this will at least prevent the router from breaking.

src/objects/viewport.ts
pleek91, line 1: The teardown issue is because of a bug in pixi-viewport. From what i understand it tries to teardown the ticker (or stop something with the ticker) but since it was using the applications ticker (and that ticker was already destroyed by the application itself) the viewport errored when it was destroyed. 

Open issue [here](https://github.com/davidfig/pixi-viewport/issues/438)

We added the application ticker here when we were building guides. But I think just making the guides render sync fixed the lag issue we were seeing. Because the guides are very smooth even without this. Everything I tested that used a ticker seemed fine. So I think this is safe to remove and it fixes the teardown issue.

src/textures/cap.ts
pleek91, line 1: Just implementing the new cache object

src/factories/guide.ts
pleek91, line 1: I've noticed quite a few issues with guides not showing up when they should. Which seems to be related to culling because if I don't cull guides then they seem to always appear when when they should. There are only 30 guides and we're updating their position when needed. So we don't need to cull from a visual perspective and the performance savings is minimal. So opting to remove guide culling.

src/factories/nodeBar.ts
```diff
@@ -25,7 +26,8 @@ export async function nodeBarFactory() {
     const { background = '#fff' } = config.styles.node(node)
     const { nodeHeight: height, nodeRadius: radius } = config.styles
     const width = getTotalWidth(node, radius)
-    const capRight = node.state_type !== 'RUNNING'
+
+    const capRight = node.state_type !== 'RUNNING' || settings.isDependency()
```
pleek91, line 30: When using dependency view all bars should have right caps

src/objects/config.ts
```diff
@@ -9,7 +9,7 @@ let config: RequiredGraphConfig | null = null
 const defaults: Omit<RequiredGraphConfig, 'runId' | 'fetch'> = {
   animationDuration: 500,
   disableAnimationsThreshold: 500,
-  disableEdgesThreshold: 100,
+  disableEdgesThreshold: 500,
```
pleek91, line 12: Not really a bug fix but we can easily render more edges and a pretty small flow I was looking at had more than 100 edges. So bumping this up to match the animation threshold.

brandonreid (approved):
Not sure why the guide culling was removed? Otherwise looks good!

znicholasbrown (approved):

PR #316, opened 2023-11-06
rename layouts and add Sorted Run Duration

brandonreid:
https://github.com/PrefectHQ/graphs/assets/6776415/d90d4403-95ea-4282-853f-bc9b31461a89

### DIFF
--- a/src/components/RunGraphSettings.vue
+++ b/src/components/RunGraphSettings.vue
@@ -14,14 +14,14 @@
     </template>
 
     <p-overflow-menu class="run-graph-settings__menu">
-      <p-label label="View">
-        <p-radio-group v-model="horizontal" :options="horizontalOptions">
+      <p-label label="Layout">
+        <p-radio-group v-model="selectedLayoutOption" :options="layoutOptions">
           <template #label="{ option }">
             {{ option.label }}
           </template>
         </p-radio-group>
       </p-label>
-      <template v-if="layout.isTemporal()">
+      <template v-if="layout.isTemporal() || layout.isLeftAligned()">
         <p-label label="Scaling" class="mt-4">
           <div class="flex items-center gap-2">
             <p-button title="Decrease scale (-)" small icon="MinusIcon" @click="decreaseScale" />
@@ -32,14 +32,6 @@
           </div>
         </p-label>
       </template>
-      <p-divider />
-      <p-label label="Layout">
-        <p-radio-group v-model="vertical" :options="verticalOptions">
-          <template #label="{ option }">
-            {{ option.label }}
-          </template>
-        </p-radio-group>
-      </p-label>
       <p-divider />
       <p-checkbox v-model="hideEdges" label="Hide dependency arrows" />
     </p-overflow-menu>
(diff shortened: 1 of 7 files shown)


src/workers/layouts/horizontal.ts
```diff
@@ -54,3 +58,18 @@ function getHorizontalTimeLayout({ data, horizontalSettings }: ClientLayoutMessa
   return layout
 }
 
+function getHorizontalLeftAlignedLayout({ data, horizontalSettings }: ClientLayoutMessage): HorizontalLayout {
+  const scale = horizontalScaleFactory(horizontalSettings)
+  const layout: HorizontalLayout = new Map()
+
+  const x = scale(data.nodes.entries().next().value[1].start_time ?? new Date())
```
pleek91, line 65: This just gets the start time of the first entry right?

pleek91:
@brandonreid This appears to remove the ability to do dependency + waterfall. Is that intentional?

src/components/RunGraphSettings.vue
```diff
@@ -62,43 +54,33 @@
...
-      return layout.horizontal
-    },
-    set(value) {
-      setHorizontalMode(value)
-    },
-  })
+  type LayoutOption = `${HorizontalMode}_${VerticalMode}`
 
-  const verticalOptions: Option<VerticalMode>[] = [
+  const layoutOptions: Option<LayoutOption>[] = [
```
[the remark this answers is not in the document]
brandonreid: That's a good call, I'll throw the names out there, maybe a Will or others have better ideas on these.

brandonreid:
> @brandonreid This appears to remove the ability to do dependency + waterfall. Is that intentional?

@pleek91 Yeah, open to pushback there, but the dependency waterfall seemed like an irrelevant way to view things.. I guess if you care about run order?

pleek91:
> > @brandonreid This appears to remove the ability to do dependency + waterfall. Is that intentional?
> 
> @pleek91 Yeah, open to pushback there, but the dependency waterfall seemed like an irrelevant way to view things.. I guess if you care about run order?

I'd personal like to not have waterfall at all. But its the fastest layout which I think is the only reason its there. For performance on large graphs.

src/workers/layouts/horizontal.ts
```diff
@@ -54,3 +58,18 @@ function getHorizontalTimeLayout({ data, horizontalSettings }: ClientLayoutMessa
   return layout
 }
 
+function getHorizontalLeftAlignedLayout({ data, horizontalSettings }: ClientLayoutMessage): HorizontalLayout {
+  const scale = horizontalScaleFactory(horizontalSettings)
+  const layout: HorizontalLayout = new Map()
+
+  const firstEntryStartTime = data.nodes.entries().next().value[1].start_time
```
pleek91, line 65: Could this just use the start time of the run itself? `data.start_time`

brandonreid: ugg... totally. haha updating

pleek91 (approved):

PR #322, opened 2023-11-08
Update node axis positions on layout update

brandonreid:
When a new layout message comes in, there's a chance that nodes have shifted into a different row or column. This PR introduces an update check loop when the layout is updated to reset those positions.

### DIFF
--- a/src/factories/nodes.ts
+++ b/src/factories/nodes.ts
@@ -327,8 +327,8 @@ export async function nodesContainerFactory() {
         continue
       }
 
-      rows.updateOffsetAxisParameters({ nodeId, axis: nodeLayout.y })
-      columns.updateOffsetAxisParameters({ nodeId, axis: nodeLayout.column })
+      rows.updateOffsetAxis({ nodeId, axis: nodeLayout.y })
+      columns.updateOffsetAxis({ nodeId, axis: nodeLayout.column })
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/objects/viewport.ts
```diff
@@ -85,7 +85,7 @@ export async function centerViewport({ animate }: CenterViewportParameters = {})
       x: x + width / 2,
       y: y + height / 2 - guidesOffset,
     },
-    scale,
+    scale: Math.min(scale, 1),
```
brandonreid, line 88: This will stop showing in this diff once the [upstream PR](https://github.com/PrefectHQ/graphs/pull/321/files) is rebased.

src/factories/nodes.ts
```diff
@@ -46,11 +46,14 @@ export async function nodesContainerFactory() {
...
-      columns.clear()
-      await render(runData!)
+  emitter.on('layoutUpdated', () => {
+    rows.clear()
+    columns.clear()
+  })
+
+  emitter.on('layoutSettingsUpdated', () => {
+    if (runData && Boolean(container.parent)) {
+      render(runData)
```
brandonreid, line 56: Not exactly sure why it's showing this in the diff, it's the same change as the upstream branch.

pleek91 (approved):

PR #350, opened 2023-12-04
Fix degraded performance from running nodes

pleek91:
# Description
As nodes run they render on the application ticker. Each time a node renders it requires a number of textures. Textures which prior to this PR were cached via subscriptions. This caused a problem because even though the graph is a single component, each time a node rendered it would get a unique subscription. So even though the texture was indeed cached. The number of subscriptions would grow by `number of running nodes * number of ticks * number of textures`. 

This caused the number of subscriptions to sky rocket as nodes ran. Slowing down each consecutive node render as the cost for creating a subscription got higher and higher. FPS would degrade slowly and if the nodes ran long enough subscriptions would exceed the stack size. 

Switching from using subscriptions to using a very simple cache with single references per cache fixes this. With this fix I can no longer see the FPS drop even after leaving nodes running for long periods of time.

### DIFF
--- a/src/objects/cache.ts
+++ b/src/objects/cache.ts
@@ -1,22 +1,28 @@
-import { Action, ActionArguments, UseSubscription, useSubscription } from '@prefecthq/vue-compositions'
+import { Action } from '@prefecthq/vue-compositions'
 
-let subscriptions: UseSubscription<() => unknown>[] = []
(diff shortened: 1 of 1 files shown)


znicholasbrown (approved):
great find 👍🏻

PR #356, opened 2023-12-12
Fix default multiplier being infinite because of 0 second runs

pleek91:
# Description
if the difference between the start_time and end_time of the run is less than 1 second we end up dividing by 0 which produces `INFINITY`. That value get's used for the horizontal scale which messes everything up and produces an empty graph.


zhen0 (approved):
Nice catch and thank you! 🙌

PR #387, opened 2024-01-31
add artifacts fetcher

brandonreid:
This PR extends the config to include additional fetchers, in this case one for "artifacts" has been added. Keeping the fetchers separate, rather than combining the data or updating the run graph API, keeps things flexible as additional layers are added in the Flow Run Graph layers raft. The `dataFactory` is still solely responsible for retrieving all data, so that additional layers are accessible to nodes right away.

### DIFF
--- a/src/models/artifact.ts
+++ b/src/models/artifact.ts
@@ -1,0 +1,16 @@
+export const artifactTypes = [
+  'result',
+  'markdown',
+  'table',
+  'unknown',
+] as const
+
+export type ArtifactType = typeof artifactTypes[number]
+
+export type Artifact = {
+  id: string,
+  flowRunId: string | null,
+  taskRunId: string | null,
+  created: Date,
+  type: ArtifactType,
+}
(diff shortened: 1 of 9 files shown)


src/models/artifact.ts
```diff
@@ -0,0 +1,16 @@
+export const artifactTypes = [
+  'result',
+  'markdown',
+  'table',
+  'unknown',
+] as const
+
+export type ArtifactType = typeof artifactTypes[number]
+
+export type Artifact = {
```
brandonreid, line 10: This isn't a full copy of the `Artifact` type from the downstream ui-library repo, it only includes fields I think we'll need. We do this with state types as well. Open to ideas on a better strategy, but this will keep the train moving for now.

pleek91: Yeah this makes sense to me. We might also want to have an api endpoint we can hit that returns a striped down version of the artifacts. That way we're not sending giant data tables to the browser just to show the title in a box.

pleek91 (approved):

PR #388, opened 2024-02-01
Render artifact nodes

brandonreid:
This PR adds the artifactFactory and uses artifact data to render them on nodes.

image

The `nodeContainerFactory` has been updated to intake the whole data bundle so it can manage nodes and artifacts for it’s children. Since nodes subscribe to a reference of their data to running state changes, otherwise the node render function isn't called again outside of the node, I've opted to have the parent `nodesContainerFactory` responsible for calling a `renderArtifacts()` function on any nodes that have them when data is updated. Certainly open to feedback on how this strategy is going.


src/models/dataBundle.ts
pleek91, line 1: This used to only have one type of date. So `RunGraphData` made sense. But now we're wrapping that in `DataBundle` and now we've got two non descriptive "data" models. I'd recommend updating `RunGraphData` to be
```
export type RunGraphData = {
  nodes: NodesData,
  artifacts: Artifact[],
}
```

That help with readability so we don't have quite so many things called "data".

src/factories/nodes.ts
```diff
@@ -64,18 +65,20 @@ export async function nodesContainerFactory() {
...
 
-    runData = data
+    dataBundle = data
     nodesLayout = null
 
     await Promise.all([
-      createNodes(data),
-      createEdges(data),
+      createNodes(),
+      createEdges(),
```
pleek91, line 76: By not passing data into these methods now the methods have to check if data is defined. So I'd switch that back (because we know the data exists here where they are called) and just pass in the specific piece of data each method needs.

src/factories/nodes.ts
```diff
@@ -64,18 +65,20 @@ export async function nodesContainerFactory() {
...
     ])
 
-    getLayout(data)
+    await createArtifacts()
```
pleek91, line 79: Why are we rendering artifacts here in nodes? I expected to see each nodes artifacts passed to it and then each node can be responsible for rendering its own artifacts. That way it can render things relative to itself. It looks like you're already calling `node.renderArtifacts`. So I think it would make sense to just pass the artifacts into the node when rendering the node.

src/factories/nodes.ts
```diff
@@ -152,7 +167,55 @@ export async function nodesContainerFactory() {
...
+
+      const group = grouped.get(artifact.taskRunId) ?? []
+
+      group.push(artifact)
+
+      grouped.set(artifact.taskRunId, group)
+    }
+
+    return grouped
+  }
```
pleek91, line 210: This grouping logic could be abstracted into a utility in a separate file. That way we can keep this file smaller.

pleek91 (commented):
Comments are for this implementation. But reading this and seeing your message in slack about an endpoint for getting stripped down artifacts made me think that we should just see if we can update the existing purpose built endpoint to also return artifacts. Basically update RunGraphNode to have an `artifacts` property that is the artifacts for that specific node.

src/factories/nodes.ts
brandonreid: Yeah.. I think you're right on adding artifacts to the node's data itself.

The reason for calling the render method from this level is just that the individual node wouldn't know when artifacts have been added to the artifacts array, since they're only subscribed to their individual node data. I thought about some separate data polling service that nodes can subscribe to directly, that feels wack though because then you have a lot of nodes subscribing to an updated event and seeing if they have any.

pleek91: Yeah I think adding artifacts to the date itself is the way to go here if that's not to bad from an api perspective. Otherwise I think that the node render function can just accept the artifacts as well.

brandonreid:
Closing in favor of a different approach: https://github.com/PrefectHQ/graphs/pull/389

PR #389, opened 2024-02-02
add artifacts to nodes

brandonreid:
This is take 2 on adding artifacts to nodes, this time by adding artifacts to node data.

image


src/factories/artifact.ts
```diff
@@ -0,0 +1,77 @@
...
+import { Artifact } from '@/models/artifact'
+import { waitForConfig } from '@/objects/config'
+import { waitForCull } from '@/objects/culling'
+
+export type ArtifactFactory = Awaited<ReturnType<typeof artifactFactory>>
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export async function artifactFactory(artifact: Artifact) {
+  const element = new Container()
+  const content = new Container()
```
brandonreid, line 13: The `content` Container holds the label and eventually will hold the icon as well. That way the combined dimensions of content can be used for sizing the bar.

src/factories/node.ts
```diff
@@ -62,7 +65,10 @@ export async function nodeContainerFactory(node: RunGraphNode) {
 
     cacheKey = currentCacheKey
 
-    await renderNode(node)
+    await Promise.all([
+      renderNode(node),
+      createArtifacts(node.artifacts),
```
pleek91, line 70: You'll need to add something about the artifacts to the cache key otherwise artifacts updating won't cause a render.

pleek91 (approved):
This looks good, one comment about the node cache key but otherwise good to go I think.

PR #390, opened 2024-02-02
add cache key for artifacts, update demo data

brandonreid:
Adds an artifacts count to the cache key for nodes and updates all the demo data to include the required `artifacts: []` array.


pleek91 (approved):
I'd probably leave artifacts as a non required property and just default the length check to 0. Otherwise you won't be able to merge your work until the api get's updated. But in generally using length for the cache key should work well

brandonreid:
> I'd probably leave artifacts as a non required property and just default the length check to 0. Otherwise you won't be able to merge your work until the api get's updated. But in generally using length for the cache key should work well

Ugg.. that's a great call actually. Adding this empty array to every object makes me wonder if the API should consider it optional as well. I'll redo this PR, haha. Take 2 season.

brandonreid:
For posterity, take 2: https://github.com/PrefectHQ/graphs/pull/391

PR #393, opened 2024-02-05
Icons on artifacts

brandonreid:
Adds an iconFactory and icon texture cache, then uses them on artifact nodes.

image


src/models/artifact.ts
```diff
@@ -13,4 +15,11 @@ export type Artifact = {
...
-}
\ No newline at end of file
+}
+
+export const artifactTypeIconMap = {
+  markdown: 'ArtifactMarkdown',
+  table: 'ArtifactTable',
+  result: 'ArtifactResult',
+  unknown: 'Artifact',
+} as Record<ArtifactType, IconName>
```
pleek91, line 25: This will give you better type safety
```suggestion
} as const satisfies Record<ArtifactType, IconName>
```

pleek91 (approved):
Nice

PR #395, opened 2024-02-07
Place artifacts

brandonreid:
Places the artifact(s) nicely on the nodes.

image

When the node is smaller than the width of the artifacts:
image


src/factories/artifact.ts
```diff
@@ -24,8 +22,6 @@ export async function artifactFactory(artifact: Artifact) {
   content.addChild(label)
   element.addChild(content)
 
-  cull.add(element)
```
brandonreid, line 27: Since artifacts are attached to nodes, culling happens on that level. It's important to not happen here since resizing the x-axis recalculates node widths, and if we cull here then they'll have a width of 0 and won't be placed properly on the newly resized node if it's out of view.

pleek91 (approved):

PR #396, opened 2024-02-07
add artifact selection

brandonreid:
When artifacts are clicked, an event is emitted so that a drawer can be rendered with artifact details.

https://github.com/PrefectHQ/graphs/assets/6776415/e0ea4864-218c-4050-9fcd-e511c0c46c7a

### DIFF
--- a/src/components/RunGraph.vue
+++ b/src/components/RunGraph.vue
@@ -13,6 +13,7 @@
   import { useKeyDown } from '@prefecthq/vue-compositions'
   import { computed, onBeforeUnmount, onMounted, ref, watch } from 'vue'
   import RunGraphSettings from '@/components/RunGraphSettings.vue'
+  import { InspectedArtifact } from '@/models'
   import { RunGraphProps } from '@/models/RunGraph'
   import { NodeSelection } from '@/models/selection'
   import { ViewportDateRange } from '@/models/viewport'
@@ -32,6 +33,7 @@
     (event: 'update:viewport', value: ViewportDateRange): void,
     (event: 'update:fullscreen', value: boolean): void,
     (event: 'update:selected', value: NodeSelection | null): void,
+    (event: 'inspectArtifact', value: InspectedArtifact): void,
   }>()
 
   const stage = ref<HTMLDivElement>()
@@ -49,7 +51,11 @@
 
   watch(() => props.selected, selected => selectNode(selected))
 
-  emitter.on('nodeSelected', nodeId => emit('update:selected', nodeId))
+  emitter.on('nodeSelected', selection => emit('update:selected', selection))
+
+  emitter.on('inspectArtifact', artifact => {
+    emit('inspectArtifact', artifact)
+  })
 
   watch(() => props.viewport, viewport => updateViewportFromDateRange(viewport))
 
--- a/src/models/selection.ts
+++ b/src/models/selection.ts
@@ -3,4 +3,5 @@ import { RunGraphNode } from '@/models/RunGraph'
 export type NodeSelection = {
   id: string,
   kind: RunGraphNode['kind'],
+  artifacts?: RunGraphNode['artifacts'],
 }
--- a/src/services/inspection.ts
+++ b/src/services/inspection.ts
@@ -1,0 +1,13 @@
+import { Artifact } from '@/models'
+import { emitter } from '@/objects/events'
+import { getSelected, selectNode } from '@/objects/selection'
+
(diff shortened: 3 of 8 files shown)


src/services/inspection.ts
```diff
@@ -0,0 +1,13 @@
+import { Artifact } from '@/models'
+import { emitter } from '@/objects/events'
+import { getSelected, selectNode } from '@/objects/selection'
+
+export function inspectArtifact(artifact: Artifact): void {
+  const selectedNode = getSelected()
+
+  if (selectedNode && !selectedNode.artifacts?.includes(artifact)) {
```
brandonreid, line 8: When a node is selected, clicking an un-highlighted node to select/highlight it works pretty intuitively. It was weird to have a path highlighted, then click an artifact on an un-highlighted node and have the artifact drawer appear, but not affect the highlighting at all. I opted here to simply deselect at that point.

src/components/RunGraph.vue
```diff
@@ -49,7 +51,11 @@
 
   watch(() => props.selected, selected => selectNode(selected))
 
-  emitter.on('nodeSelected', nodeId => emit('update:selected', nodeId))
+  emitter.on('nodeSelected', selection => emit('update:selected', selection))
+
+  emitter.on('inspectArtifact', artifact => {
+    emit('inspectArtifact', artifact)
+  })
```
pleek91, line 58: it be nice to get some consistent naming and terminology going here. We already have `NodeSelection` as a possible value for `selected`. Could we add an `ArtifactSelection` type to that and reuse the same event?

src/models/selection.ts
```diff
@@ -3,4 +3,5 @@ import { RunGraphNode } from '@/models/RunGraph'
 export type NodeSelection = {
   id: string,
   kind: RunGraphNode['kind'],
+  artifacts?: RunGraphNode['artifacts'],
```
pleek91, line 6: Why do we need to emit all the artifacts?

src/components/RunGraph.vue
brandonreid: Yeah so, just trying to think ahead for when we have a series of things that emit a value for showing pop overs (or a drawer in the case of artifacts), it seems like these should be distinct from selections which have a relationship with the selectedNode state. I could make this `inspect` emitter more generic, then include a `type` so the UI knows what popup and content to show. Does it make sense to manage `type` rather than just having separate events? Then the data passed in the event can be what's best for the type. Like I think I'll actually end up including the artifact `key` (name) as well so that a link can be seen quickly in the flyout while the rest of the (possibly big) artifact loads. Thoughts on that? The picture could get clearer as those other layers are added on too.

src/models/selection.ts
brandonreid: It's actually being emitted anyway, since the whole node is emitted. This is just a collection of properties that are consistently emitted between flow and task selections. Including it in the type here means that when I call `getSelection` I can check if the inspected artifact is associated with the selected node.

src/components/RunGraph.vue
brandonreid: You've got me thinking artifacts should appear selected when clicked now.. Possibly other "inspectable" things as well. Probably a cleaner paradigm for sure. Will mess with that.

brandonreid: Yeah it get's all mixed up with the highlighting logic, I'm still thinking they need to be separated.

pleek91: what do you mean gets mixed up? Because a node needs to be highlighted at the same time as a artifact? 

I'm thinking selected would be a discriminating union so no matter what is selected we'd know what type of thing it is and any necessary data about it. You can only select/click on one thing at a time right? So having multiple events for selection doesn't seem like a good pattern to me.

brandonreid: Nodes subscribe to selection updates and dim themselves if they're not selected. I suppose the consumer can set the selection to null if it's not a node, but that would prevent showing any sort of selected state on the artifact so I'm not sure what the point would be then. The interactions differ in that nodes are selectable, artifacts, events, and automations are "inspectable", so you click them to trigger a popover, but you're not actually selecting them the same way as you would a node, where the selection is persisted in state and updates the view. I do still totally think you're on to something though. I'll throw something on the calendar for a quick sync discussion on it.

brandonreid:
Closing in favor of https://github.com/PrefectHQ/graphs/pull/399

PR #398, opened 2024-02-08
Flow run level artifacts

brandonreid:
This PR introduces flow run level artifacts and implements clustering logic for when they collide:

https://github.com/PrefectHQ/graphs/assets/6776415/fb434f5c-c593-4463-9253-bebf0a97f9bd

Things to note:
- z-index constants have been added so that the artifacts are displayed above the viewport content
- the `artifactCluster` factory is pretty similar to the `artifact` one, but it seemed cleaner to separate them rather than add conditional logic, open to push back on that of course.
- the `flowRunArtifact**s**` factory has the clustering logic. It's debounced pretty aggressively, I felt like in real world scenarios hyper-responsiveness isn't worth the performance cost.


src/factories/icon.ts
```diff
@@ -4,11 +4,13 @@ import { waitForIconCull } from '@/objects/culling'
 import { getIconTexture } from '@/textures/icon'
 
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
-export async function iconFactory() {
+export async function iconFactory(disableCulling?: boolean) {
```
pleek91, line 7: Can you make this an object so we aren't passing plain booleans? Reads better as `iconFactory({ cull: false })` or something like that. Same for label factory.

src/factories/artifactCluster.ts
```diff
@@ -0,0 +1,160 @@
...
+import { waitForScale } from '@/objects/scale'
+
+export type ArtifactClusterFactory = Awaited<ReturnType<typeof artifactClusterFactory>>
+
+type ArtifactClusterFactoryRenderProps = {
+  ids: string[],
+  date: Date,
+}
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
pleek91, line 18: we really need to just disable this rule for this whole project...

src/factories/artifactCluster.ts
```diff
@@ -0,0 +1,160 @@
...
+  element.addChild(bar)
+
+  content.addChild(icon)
+  content.addChild(label)
+  element.addChild(content)
+
+  emitter.on('scaleUpdated', updated => scale = updated)
+
+  application.ticker.add(() => {
+    updatePosition()
```
pleek91, line 43: do you know if this is running non stop? And if so can we do something to only run this when the viewport has changed?

src/factories/flowRunArtifact.ts
```diff
@@ -0,0 +1,46 @@
...
+  const application = await waitForApplication()
+  const viewport = await waitForViewport()
+  let scale = await waitForScale()
+
+  const { element, render: renderArtifact } = await artifactFactory(artifact, true)
+
+  emitter.on('scaleUpdated', updated => scale = updated)
+
+  application.ticker.add(() => {
+    updatePosition()
```
pleek91, line 23: Same here. Be nice to not constantly  be setting the position to the same value if the viewport hasn't actually changed.

src/factories/flowRunArtifacts.ts
pleek91, line 1: We should chat through some ideas for collision detection. I feel like we could do this more efficiently but I'm not totally sure I understand what's happening here.

pleek91 (approved):

src/factories/flowRunArtifacts.ts
brandonreid: Yeah it would be great to pair on that a bit, especially since I'll have another need for it with events. I'll send a calendar invite your way. Since you're out early next week I'll probably resolve the other feedback and merge this for now. Appreciate the reviews!

pleek91: yep sounds good!

src/factories/artifactCluster.ts
brandonreid: (these are mostly notes for myself)

I did try and line this up with viewport changes based on the star/end time bindings, but it fell out of sync with guides and got funky, probably because those are debounced or something.

I don't anticipate many of these on the root flow level, but I'll take a second look though and see if something else can be used. I was going to look at assigning the application ticker to the viewport so they're both using the same one, that might help keep things in sync when listening to viewport events as well.

brandonreid: I've got something working, will make a separate PR for it.

PR #399, opened 2024-02-09
Artifact selection 2

brandonreid:
Take 2 on artifact selection. [Take 1](https://github.com/PrefectHQ/graphs/pull/396) handled it with separated selection events, this approach combines node and artifact selection with one event, keeping things simple and enforcing a paradigm of "one thing selected at a time on the graph".

This also enables a simple way to show selected artifacts. You'll see in the PR that the selected border from nodes has been abstracted for use on artifacts as well.

https://github.com/PrefectHQ/graphs/assets/6776415/c5229f12-0628-4ceb-9f26-4646a695c74c


billpalombi:
This approach definitely looks better to me.

pleek91 (approved):
This looks great. Thanks for thinking through this with me!

PR #406, opened 2024-02-13
use viewport moved event instead of application tickers

brandonreid:
This PR adds a new `viewportMoved` event to the even bus. This is preferred in place of application tickers for things that exist on the stage but aren't rendered in the viewport (guides, root flow run layers).

This loom explains it in a bit more detail:
[Private Video - Watch Video](https://www.loom.com/share/1421e307f2d847d8892455bd2a307122)
(at the end of the video I say I messed around with "culling" but I meant "debouncing")


src/objects/viewport.ts
```diff
@@ -40,6 +40,16 @@ export async function startViewport(): Promise<void> {
...
 
+  // Both the `moved` and `moved-end` events are required for a reliable `viewportMoved` event call that
+  // includes the final positions for both scroll wheel and touch pad interactions.
+  viewport
+    .on('moved', () => {
+      emitter.emit('viewportMoved', null)
+    })
+    .on('moved-end', () => {
+      emitter.emit('viewportMoved', null)
+    })
```
brandonreid, line 51: Something I didn't mention in the video. I did try and pass the event returned from the `moved` and `moved-end` callbacks, which include a reference to the viewport, for use in the update calls thinking maybe they return a non-delayed value. Unfortunately I got the same result as simply referencing the `viewport` object directly.

pleek91 (approved):
Nice! Yeah this is much better.

PR #407, opened 2024-02-13
Selectable root run level artifacts

brandonreid:
Adds the ability to select root run level artifacts and artifact clusters.

https://github.com/PrefectHQ/graphs/assets/6776415/bcbd2af2-f6a0-4377-8b39-e911ba060c41

### DIFF
--- a/src/factories/artifactCluster.ts
+++ b/src/factories/artifactCluster.ts
@@ -1,51 +1,21 @@
-import { BitmapText, Container } from 'pixi.js'
-import { DEFAULT_ROOT_ARTIFACT_BOTTOM_OFFSET } from '@/consts'
-import { barFactory } from '@/factories/bar'
-import { iconFactory } from '@/factories/icon'
-import { nodeLabelFactory } from '@/factories/label'
-import { selectedBorderFactory } from '@/factories/selectedBorder'
-import { waitForApplication, waitForViewport } from '@/objects'
-import { waitForConfig } from '@/objects/config'
+import { artifactNodeFactory } from '@/factories/artifactNode'
 import { emitter } from '@/objects/events'
-import { waitForScale } from '@/objects/scale'
 import { isSelected, selectItem } from '@/objects/selection'
 
 export type ArtifactClusterFactory = Awaited<ReturnType<typeof artifactClusterFactory>>
 
-type ArtifactClusterFactoryRenderProps = {
+export type ArtifactClusterFactoryRenderProps = {
   ids: string[],
   date: Date,
 }
 
-type ArtifactClusterFactoryRenderBgProps = {
-  width: number,
-  height: number,
-}
-
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
 export async function artifactClusterFactory() {
-  const application = await waitForApplication()
-  const viewport = await waitForViewport()
-  const config = await waitForConfig()
-  let scale = await waitForScale()
-
-  const element = new Container()
-  const content = new Container()
-  const { element: icon, render: renderIcon } = await iconFactory({ cullAtZoomThreshold: false })
-  const { element: label, render: renderLabel } = await nodeLabelFactory({ cullAtZoomThreshold: false })
-  const { element: bar, render: renderBar } = await barFactory()
-  const { element: border, render: renderBorder } = await selectedBorderFactory()
+  const { element, render: renderArtifactNode } = await artifactNodeFactory({ cullAtZoomThreshold: false })
 
   let currentDate: Date | null = null
   let currentIds: string[] = []
-  let isClusterSelected = false
-
-  element.addChild(bar)
-  element.addChild(border)
-
-  content.addChild(icon)
-  content.addChild(label)
-  element.addChild(content)
+  let selected = false
 
   element.eventMode = 'static'
   element.cursor = 'pointer'
@@ -54,14 +24,11 @@ export async function artifactClusterFactory() {
     event.stopPropagation()
     selectItem({ kind: 'artifactCluster', ids: currentIds })
   })
-
-  emitter.on('scaleUpdated', updated => scale = updated)
-  emitter.on('viewportMoved', () => updatePosition())
   emitter.on('itemSelected', () => {
     const isCurrentlySelected = isSelected(currentIds)
 
-    if (isCurrentlySelected !== isClusterSelected && currentDate) {
-      isClusterSelected = isCurrentlySelected
+    if (isCurrentlySelected !== selected && currentDate) {
+      selected = isCurrentlySelected
       render({ ids: currentIds, date: currentDate })
     }
   })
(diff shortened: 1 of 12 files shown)


src/factories/artifactCluster.ts
```diff
@@ -58,12 +89,18 @@ export async function artifactClusterFactory() {
...
 
-    await renderBg()
+    const width = artifactPaddingLeft + content.width + artifactPaddingRight
+    const height = content.height + artifactPaddingY * 2
 
-    element.visible = true
+    await Promise.all([
+      renderBg({ width, height }),
+      renderBorder({ item: currentIds, width, height }),
+    ])
```
brandonreid, line 98: Here we're sending details so the `renderBorder` factory will have the items needed to decide whether to render the selected border or not.

src/factories/artifactCluster.ts
```diff
@@ -132,11 +166,18 @@ export async function artifactClusterFactory() {
       return
     }
 
-    const xPos = scale(currentDate) * viewport.scale._x + viewport.worldTransform.tx
-    const x = xPos - element.width / 2
-    const y = application.screen.height - element.height - DEFAULT_ROOT_ARTIFACT_BOTTOM_OFFSET
+    let selectedOffset = 0
```
brandonreid, line 169: Here, we're doing some fu to place the cluster consistently, accounting for whether it's selected and the blue border is visible or not.

We didn't have to worry about this with nodes on the stage because their position isn't updated constantly like it is at this level.

src/factories/flowRunArtifact.ts
```diff
@@ -31,10 +44,18 @@ export async function flowRunArtifactFactory(artifact: Artifact) {
       return
     }
 
+    let selectedOffset = 0
```
brandonreid, line 47: Again, simply tracking whether the selected border is visible or not for the artifact, then adjusting it's placement for consistency.

src/objects/selection.ts
```diff
@@ -50,7 +50,23 @@ export function selectItem(item: GraphItemSelection | null): void {
...
+  if ('ids' in item && 'ids' in selected) {
+    return item.ids.length === selected.ids.length && selected.ids.every(id => item.ids.includes(id))
+  }
+
+  if ('id' in item && 'id' in selected) {
+    return item.id === selected.id
+  }
+
+  return false
 }
```
brandonreid, line 70: Open to feedback on this. It's flexible, which is nice when using it, but may be getting hard to maintain. I anticipate future layers to have either an id, or `ids` like artifacts, so this hopefully wouldn't require any more adjustment as additional layers are introduced.

src/factories/artifactCluster.ts
pleek91: Seems like we could add the border to a different container or use a negative position rather than offsetting the entire element.

src/factories/flowRunArtifact.ts
pleek91: It seems like the artifactCluster and flowRunArtifact share a lot of the same logic. I haven't dug into it but seems like they could follow a pattern similar to nodes where there's a single node factory that can be a flow run or a task run. But they share functionality.

src/models/selection.ts
```diff
@@ -1,15 +1,23 @@
...
 export type ArtifactSelection = {
   kind: 'artifact',
   id: string,
 }
 
-export type GraphItemSelection = NodeSelection | ArtifactSelection
+export type ArtifactClusterSelection = {
+  kind: 'artifactCluster',
+  ids: string[],
+}
```
pleek91, line 16: super nit but this might be a nice pattern since we'll have other things that are clustered
```suggestion
export type ArtifactsSelection = {
  kind: 'artifacts',
  ids: string[],
}
```

src/models/selection.ts
```diff
@@ -1,15 +1,23 @@
...
+
+export type GraphItemSelection =
+  | NodeSelection
+  | ArtifactSelection
+  | ArtifactClusterSelection
 
-export type SelectableItem = RunGraphNode | Artifact
\ No newline at end of file
+export type SelectableItem = RunGraphNode | Artifact | string[]
```
pleek91, line 23: What is the `string[]`? Multiple ids of artifacts? That might get confusing once we have multiple events selected as well.

src/objects/selection.ts
pleek91: I would probably expand on this just for readability and also type safety. I'd write separate `isSelectedNode`, `isSelectedArtifact` utilities and use those directly when possible. And then have `isSelected` call those individual utilities. 

Also I'd write type guards for the various selection kinds. So a type guard for `isNodeSelection`, `isArtifactSelection`, and `isArtifactsSelection` (or `isArtifactClusterSelection`) that way you can easily narrow a selection.

src/factories/artifactCluster.ts
brandonreid: I can offset the top and left to be a negative value outside the container, but unfortunately the right and bottom are assigned by width and height, so they still increase the container.

Appending the selected outline to a different container would have the drawback of needing to manage it's position and size everywhere it might be used separate from what it's surrounding.

I have updated this code so this offset accounting is only happening in one place. I could see a future where we have one focus factory that finds the selected element and surrounds it independently from all the individual implementations. I'll note that as a stretch goal for now though.

src/factories/flowRunArtifact.ts
```diff
@@ -1,45 +1,111 @@
...
+  const config = await waitForConfig()
   let scale = await waitForScale()
 
-  const { element, render: renderArtifact } = await artifactFactory(artifact, { cullAtZoomThreshold: false })
+  const { element, render: renderArtifact } = await getFactory()
 
-  emitter.on('scaleUpdated', updated => scale = updated)
+  let date: Date | null = artifact ? artifact.created : null
+  const isCluster = !artifact
+  let ids: string[] = artifact ? [artifact.id] : []
```
pleek91, line 26: This feels weird. If an artifact isn't passed in that means its a cluster?

src/factories/flowRunArtifacts.ts
```diff
@@ -143,46 +142,47 @@ export async function flowRunArtifactsFactory() {
...
       clusterNode = await getClusterNode()
     }
 
     const ids = [...prevIds, ...currentIds]
-    const date = new Date((prevDate.getTime() + currentDate.getTime()) / 2)
 
-    clusterNode.render({ ids, date })
+    const dates = ids.map(id => artifacts.get(id)?.getDate())
+    const minDate = new Date(Math.min(...dates.map(date => date?.getTime() ?? 0)))
+    const maxDate = new Date(Math.max(...dates.map(date => date?.getTime() ?? 0)))
```
pleek91, line 172: There are `min` and `max` utilities in `date-fns` that you could use here.

src/objects/selection.ts
```diff
@@ -49,8 +55,28 @@ export function selectItem(item: GraphItemSelection | null): void {
...
+    case 'task-run':
+      return isNodeSelection(selected) && selected.id === item.id
+    case 'flow-run':
+      return isNodeSelection(selected) && selected.id === item.id
+    case 'artifact':
+      return isArtifactSelection(selected) && selected.id === item.id
+    case 'artifacts':
+      return isArtifactsSelection(selected)
+        && selected.ids.length === item.ids.length
+        && selected.ids.every(id => item.ids.includes(id))
```
pleek91, line 75: nit: probably not worth the length checks here. 
```suggestion
      return isArtifactsSelection(selected) && selected.ids.every(id => item.ids.includes(id))
```

brandonreid: Yeah.. Even if you have a cluster selected, I think scaling the graph will clear the selection/popover

src/factories/flowRunArtifacts.ts
brandonreid: In [a PR](https://github.com/PrefectHQ/graphs/pull/410) into this one I actually change this up a bit, might as well review it there if you don't mind.

PR #408, opened 2024-02-13
add ability to disable artifacts

brandonreid:
Adds the ability to disable all artifacts. The option is optional, the idea being that we'll only show it if there is an artifacts count. Arguably, there could instead be a `has_artifacts` field on the graph data, open to opinions there.

image

### DIFF
--- a/src/objects/flowRunArtifacts.ts
+++ b/src/objects/flowRunArtifacts.ts
@@ -1,12 +1,18 @@
 import { flowRunArtifactsFactory } from '@/factories/flowRunArtifacts'
 import { RunGraphData } from '@/models/RunGraph'
 import { emitter } from '@/objects/events'
+import { waitForRunData } from '@/objects/nodes'
 
 export async function startFlowRunArtifacts(): Promise<void> {
+  const data = await waitForRunData()
   const { render: renderArtifacts } = await flowRunArtifactsFactory()
 
   function render(newData?: RunGraphData): void {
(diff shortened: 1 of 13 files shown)


src/objects/flowRunArtifacts.ts
```diff
@@ -1,14 +1,20 @@
 import { flowRunArtifactsFactory } from '@/factories/flowRunArtifacts'
 import { RunGraphData } from '@/models/RunGraph'
 import { emitter } from '@/objects/events'
+import { waitForRunData } from '@/objects/nodes'
 
 export async function startFlowRunArtifacts(): Promise<void> {
+  const data = await waitForRunData()
```
brandonreid, line 7: I learned the hard way that awaiting settings in flowRunArtifactsFactory was causing the render function to return after data was created/updated, so they wouldn't render initially. Now this object always checks for artifacts after data has been initialized.

pleek91:
@brandonreid I'm not sure we can completely know if a particular run graph has artifacts or not. If there are sub flow runs they might have artifacts and we won't have a count of those artifacts, and we won't know if they exist until the user expands that node. I think it would be safer and much simpler to always have the setting visible. If the user sees artifacts they can disable them. Otherwise it doesn't hurt to have the setting visible.

brandonreid:
Closed as this got too out of date. A second attempt was made in https://github.com/PrefectHQ/graphs/pull/411 and didn't include conditionally showing the hide option.

PR #409, opened 2024-02-14
Feature: artifacts layer

brandonreid:
Artifacts PR's have been reviewed and collected into this one. This PR is the full artifacts layer.


src/factories/flowRunArtifacts.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    await Promise.all(promises)
+  }
+
+  async function createArtifact(artifact: Artifact): Promise<BoundsContainer> {
+    if (artifacts.has(artifact.id)) {
+      return artifacts.get(artifact.id)!.render()
+    }
+
+    const factory = await flowRunArtifactFactory({ type: 'artifact', artifact }) as ArtifactFactory
```
pleek91, line 65: Why do we need an `as` here?

src/factories/flowRunArtifacts.ts
```diff
@@ -0,0 +1,233 @@
...
+    visibleItems.sort((itemA, itemB) => itemA.element.x - itemB.element.x)
+
+    await checkCollisions(visibleItems, availableClusterNodes)
+
+    for (const cluster of availableClusterNodes) {
+      cluster.render()
+    }
+  }, DEFAULT_ROOT_ARTIFACT_COLLISION_THROTTLE)
+
+  async function checkCollisions(
```
pleek91, line 113: at some point I'd still like to see if we can abstract this clustering logic out into a separate agnostic service or utility. Perhaps when you get to events it'll make sense to do that.

pleek91 (approved):
Overall LGTM! Nicely done

src/factories/flowRunArtifacts.ts
brandonreid: I updated the type definitions in [flowRunArtifact](https://github.com/PrefectHQ/graphs/pull/409/files#diff-86da312407c4c0eb549c66a3bf11bf81d2f1b720d17b6231b80a4f9c8ee432aaR15) so these `as` assignments are no longer here (very happy to remove that need). Could I ask you to sign off on the implementation there?

pleek91: Nice, yeah that's a nice change. Only thought is does it work to add that new factory type to the `getFactory` function so that you don't even need the `as` on the return? If not no worries, improvement either way.

brandonreid: It appears typescript can't narrow the type of T within the if statement there. I messed around with different ways of doing it, including a type guard for the condition. Learned some stuff about typescript getting this far though.

PR #411, opened 2024-02-16
Selection, clustering improvements, and hiding artifacts layer

brandonreid:
This PR combines a few parallel threads that were happening in PR's and branches that were becoming a pain to manage (including [selection](https://github.com/PrefectHQ/graphs/pull/407), [clustering](https://github.com/PrefectHQ/graphs/pull/410), and [disabling](https://github.com/PrefectHQ/graphs/pull/408)).

This PR represents finalizing the artifacts layer and includes:
- Root flow run level artifact selection and improvements to selection overall.
- Some improvements to the clustering algorithm so that they don't start form the beginning every time.
- General restructuring of artifacts factories to be more DRY.
- A setting for disabling artifacts.


src/factories/flowRunArtifact.ts
brandonreid, line 1: @pleek91 I threw some time on the calendar to talk about this one. It's a weird pattern but kinda nice. Essentially a thin wrapper around two different types of factories, the only real responsibility here is updating position. Maybe that's a weird or bad pattern, maybe this file can be renamed, maybe it's fine. I feel like there's room for some typescript fu to define this in `flowRunArtifacts` as like, a FlowRunArtifactFactory->Artifact variant, idk.

package.json
```diff
@@ -64,13 +63,14 @@
   },
   "dependencies": {
     "@pixi-essentials/cull": "2.0.0",
+    "@types/lodash.throttle": "^4.1.9",
```
pleek91, line 66: Can you install this as a dev dependency

src/factories/flowRunArtifact.ts
pleek91: Yeah this seems fine to me

pleek91 (approved):

PR #414, opened 2024-02-16
Flow run states layer

brandonreid:
Adds flow run level state events to the graph. States are rendered in a way so they won't go off into infinity in cases of terminal states or schedules states with occurred dates long in the past.

https://github.com/PrefectHQ/graphs/assets/6776415/839d4500-8ce3-4fd2-822f-e6d7157379b9


src/factories/flowRunState.ts
```diff
@@ -0,0 +1,107 @@
...
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export async function flowRunStateFactory(state: StateEvent, options?: FlowRunStateFactoryOptions) {
+  const application = await waitForApplication()
+  const viewport = await waitForViewport()
+  const config = await waitForConfig()
+  let scale = await waitForScale()
+
+  const element = new Container()
+  const bar = await rectangleFactory()
+  const area = await rectangleFactory()
```
pleek91, line 31: I love that we have these primitive shape factories that we can just reuse again and again.

pleek91 (approved):
This is sweet. Code follows existing patterns so not much to think about here. LGTM!

PR #418, opened 2024-02-23
Events layer on the root flow run

brandonreid:
This PR adds the events layer on the root flow run level.

https://github.com/PrefectHQ/graphs/assets/6776415/e70e7fca-fb40-40df-a5d6-5afc5cb37418

In addition, the clustering algorithm has been abstracted from the FlowRunArtifacts factory so it can be used in both places.


src/factories/flowRunArtifacts.ts
```diff
@@ -98,112 +101,28 @@ export async function flowRunArtifactsFactory() {
   }
 
   const checkLayout = throttle(async () => {
-    const visibleItems = [...artifacts.values()]
-    const availableClusterNodes = [...clusterNodes]
-    visibleItems.sort((itemA, itemB) => itemA.element.x - itemB.element.x)
+    availableClusterNodes = [...clusterNodes]
```
pleek91, line 104: Do we need to spread into a new array here? Looks like we're just looping over and calling render

brandonreid: So `availableClusterNodes` is reset here so that as the collision check goes, it can grab a cluster or create a new one if there are none or they're all being used. When you zoom in and out, clusters can be created, then no longer needed. Rather than destroying them or creating another new one, they get reused.

pleek91 (approved):
Overall this LGTM. Nothing sticks out as a performance issue. Fetching events separately is probably the best solution here 👍

PR #424, opened 2024-02-28
rename types to be graphs specific

brandonreid:
Since downstream artifacts already use generic `Artifact` and `StateEvent` is also out of sync. This updates them to `RunGraphArtifact` and `RunGraphStateEvent`. Now in the ui-library mappers they'll be more readable and less prone to clashing.


collincchoy (approved):
fwiw i called it "GraphArtifact" on the backend also 👍 

(but over there, the api response schema was already called just "Graph")

PR #433, opened 2024-03-04
make state bars selectable

brandonreid:
Updates states to match the database and adds state selection.

https://github.com/PrefectHQ/graphs/assets/6776415/620bcc7a-6320-49d4-8020-2b101f278d4e


src/components/RunGraph.vue
brandonreid, line 1: The padding here was making it so things weren't clickable (e.g. state) just below the action buttons. Using placement instead allows for it.

pleek91 (approved):
LGTM. I did notice in the video is kinda looks like the running bar doesn't shrink back down when its not selected. But hard to tell.

brandonreid:
@pleek91 ah good eye. Updated to track the selected state and persist the height when selected. Otherwise it get's weird with the pop over div covering mouseover events.

PR #440, opened 2024-03-07
Subflow states

brandonreid:
Removes the border that was once used to distinguish subflows and renders it's state in time when expanded. The main flow does not render a "shadow" of the state for "Running", as that's where most of the nodes will be, but subflows do as it helps define the surface area of the subflows.

image

Example from the demo app, a little janky just because of the way we recursively get dummy data (not realistic):
https://github.com/PrefectHQ/graphs/assets/6776415/f5973fda-7434-4291-8b83-4af4f6029e72

Because the state shadow uses a level of alpha that's pretty low (10%), it actually stacks fairly nicely when expanding deeply nested subflows.
Screenshot 2024-03-07 at 7 20 55 AM


src/factories/nodesFlowRunState.ts
brandonreid, line 1: There's overlap here with `flowRunState.ts`, particularly with hover and selection, but it differs enough throughout that it became simpler to just manage them separately.

src/factories/runStates.ts
brandonreid, line 1: This is now used for the root flow run states as well as subflow states.

pleek91 (approved):

PR #443, opened 2024-03-07
bring back borders to subflows

brandonreid:
Adds back a border, but with a flat bottom (instead of rounded) so that the state doesn't have to accommodate the curve.

https://github.com/PrefectHQ/graphs/assets/6776415/70e55a2c-e16d-4d5c-827d-c019e5c6524a

### DIFF
--- a/src/factories/border.ts
+++ b/src/factories/border.ts
@@ -1,14 +1,25 @@
-import { ColorSource, Container, Sprite, Texture } from 'pixi.js'
-import { CornerStyle, getCornerTexture } from '@/textures/corner'
+import { ColorSource, Container, Sprite } from 'pixi.js'
+import { getCornerTexture } from '@/textures/corner'
 import { getPixelTexture } from '@/textures/pixel'
 
 export type BorderStyle = {
   width: number,
   height: number,
   stroke: number,
-  radius?: number,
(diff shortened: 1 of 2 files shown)


src/factories/border.ts
```diff
@@ -8,6 +8,7 @@ export type BorderStyle = {
   stroke: number,
   radius?: number,
   color?: ColorSource,
+  roundedTop?: boolean,
```
pleek91, line 11: Thoughts on adding `topLeftRadius`, `bottomLeftRadius`, etc rather than adding some specific to this implementation?

brandonreid: Yeah I almost did like `radius: number | [number, number, number, number]`. I still could. Just felt like the need to have one radius different would be pretty unlikely to happen for the effort. (have to create separate calls for corners and all that)

pleek91: seems likely that we might need a "roundedBottom" or something though. `roundedTop` seems overly specific to the currently implementation to belong in something that's generic to creating borders

brandonreid: Updated to support `radius: number | [number, number, number, number]`

pleek91 (approved):
Nice, very clean

PR #446, opened 2024-03-11
Bugfix: artifact cluster measurements on emit

brandonreid:
For root level artifacts, the viewport scale was being included in selection dimensions, which caused inaccurate overlay boxes from the popover.


pleek91:
@brandonreid can you explain why? In my mind it shouldn't matter if the element is a cluster or not the logic should be the same 🤔

znicholasbrown (approved):

brandonreid:
@pleek91 @znicholasbrown 

Wow, good looking out fellas. I think I meant to check if it was from a root node or not, but I actually don't need to scale the selection div at all. 😅 

Updated. I'll assume you'd approve the now very simple change.

PR #447, opened 2024-03-11
Bugfix: add locking mechanism to artifact and event creation

brandonreid:
Prevents the occasional scenario where multiple instances of the same artifact or event can be created if happening too quickly. When this does occur, artifacts/events are created but the duplicate isn't added to the artifact's map, thus isn't included in clustering logic.

I have seen this randomly occurring when loading a graph and was eventually able to reproduce with a running graph that created two artifacts at the same time. 

image


pleek91 (approved):
This is why the node factory has a cache key and doesn't create the node on each render. Not sure if that would work here or not since events and artifacts can be grouped, but I'd imagine it could be. Rather than creating the event or artifact on each render. Create it once and cache it, because an artifact or an event can't actually change. Just how many artifacts or events can change. 

Food for though. This works 👍

PR #450, opened 2024-03-11
Fix: Prefent node dimming if selecting from differing scope

brandonreid:
Currently, when selecting a subflow, the parent node is dimmed, causing all subflow nodes to also dim regardless of whether or not they contain a selection.

image

This PR makes it so that nodes aren't dimmed if the selected node doesn't exist in the current scope of nodes. Ideally, we'd find the subflow node containing the selection and highlight the path. This get's tricky with recursion, like if you select a subflow node that's 5 subflows deep. So this is a fast and improved solution, but not quite ideal. If I have time I'll pursue the ideal solution in a way that doesn't necessitate too much recursion.

After:
image

### DIFF
--- a/src/factories/nodes.ts
+++ b/src/factories/nodes.ts
@@ -344,7 +344,7 @@ export async function nodesContainerFactory() {
     const settings = await waitForSettings()
     const selected = getSelectedRunGraphNode()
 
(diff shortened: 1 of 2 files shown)


pleek91 (approved):
I like it, super efficient check and improves the ux

PR #455, opened 2024-03-13
Bugfix: Live running subflow updates

brandonreid:
- Fixes live running subflows
- Fixes subflow state logic that would have hidden any running state that wasn't of type "RUNNING"
- Cleans up some confusing references to `node` in `nodeFlowRun` and `nodeFlowRunState` and standardizes their use of `internalNode` so that up-to-date values are used.

Before:
https://github.com/PrefectHQ/graphs/assets/6776415/fe6f7625-c80b-464a-a3e1-5d622d94ad08

After:
https://github.com/PrefectHQ/graphs/assets/6776415/e78c4283-3354-4e52-b105-daf7b956d8d7


src/factories/nodeFlowRun.ts
```diff
@@ -37,10 +37,19 @@ export async function flowRunContainerFactory(node: RunGraphNode) {
...
-  arrowButton.zIndex = 3
+  const maxContentZIndex = Math.max(
+    DEFAULT_ROOT_EVENT_Z_INDEX,
+    DEFAULT_ROOT_FLOW_STATE_Z_INDEX,
+    DEFAULT_ROOT_ARTIFACT_Z_INDEX,
+  )
+
+  bar.zIndex = maxContentZIndex + 1
+  label.zIndex = maxContentZIndex + 2
+  arrowButton.zIndex = maxContentZIndex + 2
```
pleek91, line 52: Shouldn't we just create consts for what we expect these to be rather than making them dynamic?

src/factories/nodeFlowRunState.ts
```diff
@@ -146,18 +147,32 @@ export async function nodeFlowRunStateFactory(state: RunGraphStateEvent) {
   function getRenderStyles(): StateRectangleRenderProps {
     const { background = '#fff' } = config.styles.state(state)
 
-    let startX = scale(state.timestamp) - scale(parentStart!)
+    const parentStartX = scale(parentStart!)
```
pleek91, line 150: I feel like this `!` could produce runtime errors. Could we do something type safe here?

src/factories/node.ts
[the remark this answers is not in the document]
brandonreid: Yeah so when initiating the factory, you can provide a `node` that get's assigned to `internalNode`. When the `render()` function is called, it accepts optional new data and updates `internalNode`. Before, we had a number of functions accepting `node` as an argument, which might be valid but gets confusing given the root factory prop having the same name. So it was sort of a mess between things referencing the provided `node` on init, vs. what should be the latest data in `internalNode`. This cleans that up, same in `src/factories/nodeFlowRun.ts`

src/factories/nodeFlowRun.ts
brandonreid: Updated. In my mind I had reserved the definitions in `const` as sort of more global, whereas in this context the z-indexes only apply to this container, but realized that doesn't really matter. More readable just having `subflow` default z-indexes in the const.

pleek91 (approved):

PR #460, opened 2024-03-15
Fix: Gradient fills on nodes if zoomed screen

### LINKED ISSUE #459: Browser zoom results in gradient node fills
When using a browser zoom setting different than the default. This is two notches zoomed out:

image

The browser zoom amount may need to be accounted for in the resolution setting somehow?

brandonreid:
Resolves https://github.com/PrefectHQ/graphs/issues/459

Nodes were appearing to have gradient fills when the browser was zoomed:
image

After digging into the PIXI discord for a while, it seems that this happens when floats are provided for the resolution (e.g. `2.5` or `1.7777`). Rounding this resolution up to a whole number solves the issue and errs on the side of higher resolution rather than lower, so things still render nice and sharp.

image

### DIFF
--- a/src/objects/application.ts
+++ b/src/objects/application.ts
@@ -29,7 +29,7 @@ function createApplication(stage: HTMLDivElement): void {
     backgroundAlpha: 0,
     resizeTo: stage,
     antialias: true,
-    resolution: window.devicePixelRatio,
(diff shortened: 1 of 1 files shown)


pleek91 (approved):
Nice!

PR #469, opened 2024-03-18
Fix: Use static polling interval for events

brandonreid:
Instead of using adaptive polling for events, use a static interval defined in the graph config.

### DIFF
--- a/src/factories/eventData.ts
+++ b/src/factories/eventData.ts
@@ -1,5 +1,4 @@
 import { MaybeRefOrGetter, toValue } from 'vue'
-import { getIntervalForDataSize } from '@/factories/data'
 import { RunGraphEvent, RunGraphFetchEventsOptions } from '@/models'
 import { waitForConfig } from '@/objects/config'
 import { waitForRunData } from '@/objects/nodes'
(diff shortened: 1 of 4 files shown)


src/factories/eventData.ts
```diff
@@ -28,7 +27,7 @@ export async function eventDataFactory(
     }
 
     if (!runGraphData.end_time) {
-      interval = setTimeout(() => start(), getIntervalForDataSize(runGraphData))
+      interval = setTimeout(() => start(), config.fetchEventsInterval)
```
znicholasbrown, line 30: out of curiosity - should this be `setInterval`? The `clearInterval` on line 36 isn't doing anything. It won't throw an error because `setTimeout` and `setInterval` take the same argument (a number id)

znicholasbrown (approved):

src/factories/eventData.ts
pleek91: Its using setTimeout because its calling itself. Basically after every fetch we evaluate if we should continue fetching and if we should we schedule another fetch.

pleek91: You're right about the clearInterval it should be clearTImeout. Will update

znicholasbrown (approved):

PR #498, opened 2024-05-17
Enhancement: Progress Artifact

dylanbhughes:
This PR adds handling for displaying dynamic progress artifacts on the flow run graph.

Screenshot 2024-05-22 at 5 07 18 PM


src/factories/artifactNode.ts
```diff
@@ -12,8 +13,8 @@ type ArtifactNodeFactoryOptions = {
 type ArtifactNodeFactoryRenderOptions = {
   selected?: boolean,
   name?: string,
-  type?: ArtifactType,
-}
+  type: ArtifactType,
+} & RunGraphArtifactTypeAndData
```
pleek91, line 17: Since type exists on `RunGraphArtifactTypeAndData` It be little more readable and simpler to do this

```ts
{
  selected?: boolean,
  name?: string,
  artifact: RunGraphArtifactTypeAndData
}
```

Then you can just do `options.artifact.type` everywhere.

src/factories/artifactNode.ts
pleek91, line 1: I cannot comment on it since you didn't modify it but `async function render(options?: ArtifactNodeFactoryRenderOptions)` should probably just require the options. They're being passed in everywhere and seems like they're required in order for this to work right.

src/factories/node.ts
```diff
@@ -180,6 +180,12 @@ export async function nodeContainerFactory(node: RunGraphNode) {
   function getNodeCacheKey(nodeData: RunGraphNode): string {
     const endTime = nodeData.end_time ?? new Date()
     const artifactCount = nodeData.artifacts?.length ?? 0
+    const artifactCacheKey = nodeData.artifacts?.map(artifact => {
+      if (artifact.type === 'progress') {
+        return `${artifact.id}-${artifact.data}`
+      }
+      return ''
+    }).join()
```
pleek91, line 188: We don't really need both of these. Could you write an artifacts cache key that has a value for every artifact? I think if you just return the `artifact.id` where you're doing `return ''` that'll satisfy all artifacts. 

It also be nice if we formatted this a bit for future is if we're debugging this key. Maybe join with a separator like `,` or `|`.

pleek91 (approved):
Overall LGTM! Few suggestions but nothing blocking.

src/factories/node.ts
dylanbhughes: If the artifact type isn't progress the `data` field won't be present. I did make a change so we don't need artifactCount, though

src/factories/node.ts
```diff
@@ -179,15 +179,20 @@ export async function nodeContainerFactory(node: RunGraphNode) {
...
     const values = [
       nodeData.state_type,
       endTime.getTime(),
-      artifactCount,
       layout.horizontal,
       layout.horizontalScaleMultiplier,
       config.styles.colorMode,
-      artifactCount > 0 && settings.disableArtifacts,
+      artifactCacheKey && settings.disableArtifacts,
+      artifactCacheKey,
```
pleek91, line 195: Probably not hurting anything but I think this is what we want here. 
```suggestion
      artifactCacheKey && settings.disableArtifacts,
```

pleek91 (commented):
Realize this is already merged. No rush to fix this. Prob fine to ignore but not totally sure. Otherwise we'll end up triggering renders because artifacts changed even if artifacts are disabled.

PR #615, opened 2025-01-15
Update scripts to first run the associated command in the core project

pleek91:
# Description
In order to build the demo the core project needs to be built. So for dev and build commands we need to first dev/build the core project.

### DIFF
--- a/vue/package.json
+++ b/vue/package.json
@@ -4,7 +4,8 @@
   "main": "main.ts",
   "scripts": {
     "serve": "vite",
-    "build": "vite build"
+    "build": "cd ../core && npm run build && cd ../vue && vite build",
+    "dev": "cd ../core && npm run dev & cd ../vue && npm run serve"
   },
   "author": "Craig Harshbarger",
   "license": "ISC",


vue/package.json
```diff
@@ -4,7 +4,8 @@
   "main": "main.ts",
   "scripts": {
     "serve": "vite",
-    "build": "vite build"
+    "build": "cd ../core && npm run build && cd ../vue && vite build",
+    "dev": "cd ../core && npm run dev & cd ../vue && npm run serve"
```
znicholasbrown, line 8: have you tested this `dev` command locally? I'm surprised this would work given that `npm run dev` tends to be a process-blocking command. Does it ever get to serving the demo?

pleek91: Subtile difference between `&` and `&&` in the middle there. `&` allowing commands to run concurrently vs `&&` which is process blocking. It did work from my testing.

znicholasbrown: ah you're right I read those as the same. so you don't need the core build to complete before the serve command runs? nice

znicholasbrown (approved):

vue/package.json
pleek91: You might need to have something built first 🤔 not sure. But this did work fine for me even though the terminal output is quite messy lol.

PR #621, opened 2025-01-17
Add a react implementation

pleek91:
Add a basic react implementation and fix a bug in core that was causing the react implementation to not render correctly

image


core/.vscode/settings.json
```diff
@@ -36,8 +36,7 @@
     "vitejs"
   ],
   "files.exclude": {
-    "node_modules": true,
-    "dist": true
+    "node_modules": true
```
znicholasbrown, line 39: Is this because the core build ends up in dist in dev mode?

react/src/main.tsx
```diff
@@ -0,0 +1,9 @@
+import { StrictMode } from 'react'
+import { createRoot } from 'react-dom/client'
+import App from './App.tsx'
+
+createRoot(document.getElementById('root')!).render(
+  <StrictMode>
```
znicholasbrown, line 6: `!` 👁️ `<StrictMode>` 🤔

znicholasbrown (approved):

core/.vscode/settings.json
pleek91: This is just so I can actually see when its built because I kept wondering why it wasn't working and then I'd realize I didn't have a dist. This can probably be reverted later.

znicholasbrown: oohh ok, got it

PR #642, opened 2025-03-04
Support custom fonts

pleek91:
# Description
OSS is currently using the default fonts specified by tailwind so the graph always times out loading the webfont it specifies. Updating the config to allow for both web fonts and bitmap fonts and updating the default to match tailwind.


znicholasbrown (approved):
Don't forget to update this in downstream repos!

pleek91:
> Don't forget to update this in downstream repos!

Only the oss uiv2 is on v3 of the graph. And I don't expect us to update the vue apps. But we will need to if we ever do 👍

